PR #1, opened 2021-09-22
ONTO-59 Make sure the OP and RP support the same DID methods

maikel-maas:
- Added support to subject_identifier_types_supported, did_methods_supported and credential_formats_supported;
   - Added unit testing;


nklomp:
@maikel-maas The PR has conflicts

src/AuthenticationResponseRegistration.ts
```diff
@@ -16,5 +16,12 @@ export function createDiscoveryMetadataPayload(opts: SIOP.DiscoveryMetadataOpts)
       SIOP.SigningAlgo.EDDSA,
     ],
     subject_types_supported: opts?.subjectTypesSupported || SubjectType.PAIRWISE,
+    credential_formats_supported: opts.credential_formats_supported,
```
nklomp, line 19: Please note that opts are camelCase whilst payloads are snake_case

src/schemas/AuthenticationRequestOpts.schema.ts
```diff
@@ -219,13 +219,27 @@ export const AuthenticationRequestOptsSchema = {
...
+          ]
+        },
         "registrationBy": {
           "$ref": "#/definitions/RegistrationType"
         }
       },
       "required": [
-        "didMethodsSupported",
-        "registrationBy"
+        "credential_formats_supported",
```
nklomp: Although this is auto generated, I noticed that the didMethodsSupported is gone, which is odd

src/types/SIOP.types.ts
```diff
@@ -127,6 +127,13 @@ export interface DiscoveryMetadataOpts {
   subjectTypesSupported?: SubjectType[] | SubjectType;
   idTokenSigningAlgValuesSupported?: KeyAlgo[] | KeyAlgo;
   requestObjectSigningAlgValuesSupported?: SigningAlgo[] | SigningAlgo;
+  dids_supported: boolean;
```
nklomp: Camelcase in the opts, snake in the payload

src/types/SIOP.types.ts
```diff
@@ -150,13 +163,15 @@ export interface ResponseRegistrationOpts extends DiscoveryMetadataOpts {
 }
 
 export interface RPRegistrationMetadataOpts {
-  subjectIdentifiersSupported?: SubjectIdentifierType[] | SubjectIdentifierType;
-  didMethodsSupported: string[] | string;
+  subjectIdentifiersSupported: SubjectIdentifierType[] | SubjectIdentifierType;
+  didMethodsSupported?: string[] | string;
+  credential_formats_supported: CredentialType[] | CredentialType;
```
nklomp: camel case in opts

test/AuthenticationRequest.request.spec.ts
```diff
@@ -78,6 +78,7 @@ describe("create Request Uri should", () => {
             registration: {
                 didMethodsSupported: ['did:ethr:'],
                 subjectIdentifiersSupported: SubjectIdentifierType.DID,
+                credential_formats_supported: CredentialType.JWT,
```
nklomp: TBH i think PE support for JWT based VCs is not complete, so it might make more sense to start with json_ld type based credentials in our tests as well

src/RP.ts
```diff
@@ -93,6 +94,11 @@ export class RP {
       return this;
     }
 
+    addCredentialFormats(credentialFormat: CredentialType): RP.Builder {
```
nklomp: Singular, as it only adds a single credential format to the array

maikel-maas:
@nklomp Fixed!

src/functions/DidSiopMetadata.ts
```diff
@@ -0,0 +1,46 @@
+import { SIOP, SIOPErrors } from '../types';
+import { DiscoveryMetadataPayload, RPRegistrationMetadataPayload, SubjectIdentifierType } from '../types/SIOP.types';
+
+export function assertValidMetadata(opMetadata: DiscoveryMetadataPayload, rpMetadata: RPRegistrationMetadataPayload) {
```
nklomp, line 4: Right now we are doing verification, which obviously is nice. What I would also like to know as a user is which did methods and credentials formats etc I can actually use (so the actual intersection result). Probably makes sense to return that is a new method which returns a new interface with the supported metadata between RP and OP

nklomp (changes requested):
See remarks about camel case vs snake case.
Also I would need to have the intersection of supported metadata between RP and OP as an object, so that as an OP I can determine what did methods and credential formats I can use

src/schemas/AuthenticationRequestOpts.schema.ts
nklomp: A yupz make sense

src/types/SIOP.types.ts
maikel-maas: Fixed!

src/types/SIOP.types.ts
maikel-maas: Fixed!

test/AuthenticationRequest.request.spec.ts
maikel-maas: Fixed!

src/RP.ts
maikel-maas: Fixed!

src/functions/DidSiopMetadata.ts
maikel-maas: Fixed!

src/functions/DidSiopMetadata.ts
```diff
@@ -1,12 +1,143 @@
...
+import {
+  DiscoveryMetadataPayload,
+  RPRegistrationMetadataPayload,
+  SubjectIdentifierType,
+  SupportedTypes,
+} from '../types/SIOP.types';
 
-export function assertValidMetadata(opMetadata: DiscoveryMetadataPayload, rpMetadata: RPRegistrationMetadataPayload) {
-  verifyCredentialsFormat(rpMetadata.credential_formats_supported, opMetadata.credential_formats_supported);
+const allDidMethodsSupported = [
```
nklomp, line 9: Hmzz, DID methods are the actual part after did:<method>. So not including did:. The SIOP spec has that wrong. So lets not repeat their mistake ;)

I am also not sure whether I like this approach to begin with. Basically the RP and OP just need to accept supported methods between one another. Whether the uniresolver we use supports it or not is not really relevant. That is up to the RP and OP itself. They can always add/use a resolver of their choice. So if there is a DID method called 'shinynew' that both claim to support, then we also can expect that both can resolve them. It is then not up to us to not allow it because it is not in our internal list

src/functions/DidSiopMetadata.ts
```diff
@@ -1,12 +1,143 @@
...
+  const credentials = verifyCredentialsFormat(
+    rpMetadata.credential_formats_supported,
+    opMetadata.credential_formats_supported
+  );
   const isDid = verifySubjectIdentifiers(rpMetadata.subject_identifiers_supported);
   if (isDid && rpMetadata.did_methods_supported) {
-    verifyDidMethods(rpMetadata.did_methods_supported, opMetadata.did_methods_supported);
+    methods = verifyDidMethods(rpMetadata.did_methods_supported, opMetadata.did_methods_supported);
+  } else if (isDid && (!rpMetadata.did_methods_supported || !rpMetadata.did_methods_supported.length)) {
+    methods = verifyDidMethods(allDidMethodsSupported, opMetadata.did_methods_supported);
```
nklomp, line 138: Is the intent here that whenever the RP does not expose which DIDs it supports to accept everything, but then only the ones in our internal list above? If so we should just leave it up to the RP to try to resolve the actual DID that the OP provides. If that doesn't work, the RP will return an error anyway. Right now we are restricting to our internal list (see the 'shinynew' remark from earlier

src/functions/DidSiopMetadata.ts
```diff
@@ -31,16 +162,18 @@ function verifySubjectIdentifiers(subjectIdentifiers: SubjectIdentifierType | Su
   return subjectIdentifiers === SIOP.SubjectIdentifierType.DID;
 }
 
-function verifyDidMethods(methodsA: string[] | string, methodsB: string[] | string): void {
+function verifyDidMethods(methodsA: string[] | string, methodsB: string[] | string): Array<string> {
```
nklomp, line 165: suggest to rename to supportedDidMethods

src/functions/DidSiopMetadata.ts
```diff
@@ -31,16 +162,18 @@ function verifySubjectIdentifiers(subjectIdentifiers: SubjectIdentifierType | Su
...
   const supportedDidMethods = getIntersection(methodsA, methodsB);
```
nklomp, line 166: Suggest to do the above check for empty rp metadata first in this method. If the RP supports everything, then simply return all the results passed in for the OP

src/functions/DidSiopMetadata.ts
nklomp, line 165: Since this is always between OP and RP suggest to rename the arguments to include OP and RP respectively instead of A and B

nklomp (changes requested):
See remark about the internal did method list as well as the way we determine the intersection

src/functions/DidSiopMetadata.ts
nklomp: Also see the remark down below about using this list

src/functions/DidSiopMetadata.ts
nklomp: Will resolve this comment as I made remarks about the new methods

src/functions/DidSiopMetadata.ts
maikel-maas: Fixed!

src/functions/DidSiopMetadata.ts
maikel-maas: Fixed!

src/functions/DidSiopMetadata.ts
maikel-maas: Fixed!

src/functions/DidSiopMetadata.ts
maikel-maas: Fixed!

src/functions/DidSiopMetadata.ts
maikel-maas: Fixed!

nklomp (approved):

PR #2, opened 2021-09-28
Feature/onto 58 - added PEManger class for handling interactions between this library and @speahreon/pe-js library

sksadjad:
### Added PE-JS support to `did-auth-siop` library


src/AuthenticationRequest.ts
```diff
@@ -133,6 +135,14 @@ function createURIFromJWT(
   jwt: string
 ): SIOP.AuthenticationRequestURI {
   const schema = 'openid://';
+  const pejs: PEJS = new PEJS();
```
nklomp, line 138: Suggest to move this to a function and do the actual validation in the createJWT method instead of in the createURIFromJWT function, as that only is creating the URI as its responsibility. In createJWT the payload is created and it is the method that always will be called eventually.

src/AuthenticationResponse.ts
```diff
@@ -39,6 +41,16 @@ export default class AuthenticationResponse {
     verifiedJwt: SIOP.VerifiedAuthenticationRequestWithJWT,
     responseOpts: SIOP.AuthenticationResponseOpts
   ): Promise<SIOP.AuthenticationResponseWithJWT> {
+    const pejs: PEJS = new PEJS();
+    // bunch of ifs to make sure we need to call evaluate
+    const pd: PresentationDefinition =
+      verifiedJwt.payload.claims['id_token']['verifiable_presentations'].presentation_definition;
```
nklomp: Needs to be a bit more defensive. A definition doesn't have to be there, let alone the claims values/keys

src/AuthenticationResponse.ts
```diff
@@ -39,6 +41,16 @@ export default class AuthenticationResponse {
...
+    const result: EvaluationResults = pejs.evaluate(pd, responseOpts.vp);
+    if (result.errors.length) {
+      throw new Error(`${SIOPErrors.EVALUATE_PRSENTATION_EXCHANGE_FAILED}`);
+    }
+    const ps = pejs.submissionFrom(pd, responseOpts.vp.getVerifiableCredentials());
```
nklomp: We can start out this way for now and because of the deadline, but we really need to have a split here.

Basically we need to expose the submission requirements/definition. Then have the wallet/holder supply the the credentials to the pe-js filtering mech. Then the user selects and/or approves the actual credentials. Only after that we can generate the AuthResponse.

src/AuthenticationResponse.ts
```diff
@@ -39,6 +41,16 @@ export default class AuthenticationResponse {
...
+    responseOpts.vp.getRoot()['presentation_submission'] = ps;
```
nklomp: Hmzz, the options are being passed in by the user, I don't like we start changing the supplied options. So we need to explode the options into a new options object or something

nklomp (changes requested):
See remarks.

We certainly need to think about the interaction and splitting up on the OP side.

The user needs to be aware of what the RP is requesting, we need the filtering and actual approval of the OP to submit the VPs. Of course the actual approval is out of scope of this library because that can be implemented in many ways. But splitting it up in 2 phases is certainly something we can do. First do the checks on the AuthRequest and extract potential submission requirements etc which are validated. Then it is up to the OP software to do the actual matching and approval. The next phase for the OP class would be to actually create the AuthResponse using VPs as input from the OP software. We can do a check on the OP side at this point whether everything is valid (similar to what we also do on the RP side)

src/AuthenticationRequest.ts
```diff
@@ -135,10 +136,10 @@ function createURIFromJWT(
   jwt: string
 ): SIOP.AuthenticationRequestURI {
   const schema = 'openid://';
-  const pejs: PEJS = new PEJS();
+  const peAgent: PresentationExchangeAgent = new PresentationExchangeAgent();
   const optionalPD = extractDataFromPath(requestPayload, '$..presentation_definition');
   if (optionalPD && optionalPD.length) {
-    const validationResult: Validated | Validated[] = pejs.validateDefinition(optionalPD[0].value);
+    const validationResult: Validated | Validated[] = peAgent.validatePresentationDefinition(optionalPD[0].value);
     if (validationResult[0].message != 'ok') {
```
nklomp, line 143: Given we now have a class we control, can we move the error if the message is not 'ok' not to that method?

src/OP.ts
```diff
@@ -96,6 +102,30 @@ export class OP {
...
+    verifiableCredentials: VerifiableCredential[],
+    holderDID: string,
+    responseOpts?: {
+      nonce?: string;
+      state?: string;
+      // audience: string;
+      verification?: InternalVerification | ExternalVerification;
+    }
+  ): Promise<AuthenticationResponseWithJWT> {
+    const optionalPD = extractDataFromPath(verifiedJwt.payload, '$..presentation_definition');
```
nklomp, line 116: Can we not have the peAgent, or rather the user in the end come up with a list of VPs that we simply accept as an optional responseOpt?

Then the method to interact with to create the response stays the createJWTFromVerifiedRequest. We could make a small check there whether VPs are supplied when the request contained a PE definition.

src/OP.ts
```diff
@@ -117,6 +147,23 @@ export class OP {
     };
   }
 
+  public async selectVerifiableCredentialsForSubmission(
+    authenticationRequestPayload: AuthenticationRequestPayload,
+    verifiableCredentials: VerifiableCredential[],
+    holderDid: string
+  ): Promise<SelectResults> {
+    const optionalPD = extractDataFromPath(authenticationRequestPayload, '$..presentation_definition');
+    if (optionalPD && optionalPD.length) {
```
nklomp, line 156: Can there only be 1 definition?

src/OP.ts
```diff
@@ -117,6 +147,23 @@ export class OP {
     };
   }
 
+  public async selectVerifiableCredentialsForSubmission(
```
nklomp, line 150: I think we might want to move this method to the agent as well. Then the flow becomes to use the peAgent if VCs are being used. Do the selection and validation over there. The OP class just has the definitation and accepts VPs

src/PresentationExchangeAgent.ts
```diff
@@ -0,0 +1,40 @@
+import { EvaluationResults, PEJS, SelectResults, Validated } from '@sphereon/pe-js';
```
nklomp, line 1: Nice to have the PE Agent :) It is okay for me to have a bit of AuthReq knowledge in here, as we are otherwise basically delegating everything to pejs anyway

nklomp (changes requested):
I think we need to move small portions to the agent. The introduction of the agent is nice.

Some questions on how we handle it still

src/OP.ts
sksadjad: I've added this method here because I've considered it a method for our interface.
app (or whatever)  that uses this lib will call this method first and based on the result creates the correct presentation_submission object and sends the correct VCs

src/main/functions/DidJWT.ts
```diff
@@ -221,6 +223,12 @@ export function parseJWT(jwt: string): JWTDecoded {
   if (!payload || !header) {
     throw new Error(SIOPErrors.NO_ISS_DID);
   }
+  // Here we have to fetch the presentation_definition part of the decoded jwt
+  if (PEManager.findValidPresentationDefinition(payload)) {
+    payload['peContext'] = PresentationExchangeContext.PE;
```
nklomp, line 228: Hmzz, I don't really like this. This is adding things to an existing and signed JWT payload!

This should be done when the payload is being created by the RP

src/main/types/SIOP.types.ts
```diff
@@ -196,6 +198,11 @@ export enum ResponseContext {
   OP = 'op',
 }
 
+export enum PresentationExchangeContext {
```
nklomp, line 201: Suggest to make this something like PresentationExchangeVersion en then have NONE and V1_0 for instance. This allows us to respond to changes in the PE spec in the future

src/main/functions/DidJWT.ts
nklomp: To be honest I believe the OP side should simply parse the payload for existence of the PE definition in the payload and store that in the verfiedJwt. Reason is that we are adding logic to our implementation between the OP and RP based upon our own object. This means the OP could not work anymore with an RP using another library but conforming to SIOP v2

nklomp (changes requested):
I am not entirely clear whether the flow on the OP side is correct. See a remark where I go into what the flow should be IMO. 
Also I think we shouldn;t be adding the PEContext to the payload. That means we will not be compatible with other solutions using SIOP v2. We simply need to add the object to the verifiedJwt container object upon parsing the payload for existence of a PE definition.

nklomp (changes requested):
Small change

src/main/OP.ts
```diff
@@ -76,10 +76,13 @@ export class OP {
...
-    } else if (responseOpts && responseOpts.vp) {
-      throw new Error(`${SIOPErrors.WRONG_FUNCTION_CALL_ENDPOINT_DOESNT_ACCEPT_VCS}`);
+    const pd = PEManager.findValidPresentationDefinition(verifiedJwt.payload);
+    if (pd) {
+      if (!responseOpts && !responseOpts.vp) {
+        throw new Error(`${SIOPErrors.AUTH_REQUEST_EXPECTS_VP}`);
+      }
+      new PEManager().evaluate(pd, responseOpts.vp);
+      PEManager.validatePresentationSubmission(responseOpts.vp.getPresentationSubmission());
     }
```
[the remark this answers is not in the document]
sksadjad: already added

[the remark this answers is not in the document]
sksadjad: 87:89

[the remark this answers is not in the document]
nklomp: Sorry had my context only to the current changes. Still suggest to make it an else if (responseOpts && responseOpts.vp) ;)

nklomp (commented):
See remark

PR #3, opened 2021-09-29
Dist package fixes

sanderPostma:
Please check if this is an acceptable solution. The way how things are structed in master, since the did-jwt-fork package was added, on the implementing side I need to append /src in every import like this:
import {OP} from "@sphereon/did-auth-siop/dist/main/src"
Moving did-jwt-fork under src solves that issue but then it's mixed with the other packages. I moved those to "main" so the end package looks like this:
image


nklomp:
Before looking at it first fix the build please

nklomp:
This looks more like a problem with the index.ts or its exports. Do we have our main class properly configured?

sanderPostma:
It is set to dist/main/index.js as it should be, but because of the presence of did-jwt-fork the whole root is moved into the src folder
image
We could also change to this
  "main": "dist/main/src/index.js",
  "types": "dist/main/src/index.d.ts",
But it feels a bit clumsy to have both a dist and a src folder in the artifact path

nklomp:
Ok for now we go ahead, but we need another solutions

nklomp (approved):

PR #5, opened 2021-10-04
ONTO-58: added RP side pd vs vp evaluation

sksadjad:


src/main/types/SIOP.types.ts
```diff
@@ -91,10 +91,18 @@ export interface AuthenticationResponsePayload extends JWTPayload {
   did: string;
   registration?: DiscoveryMetadataPayload;
   registration_uri?: string;
-  vp?: VerifiablePresentation;
+  verifiable_presentations?: VerifiablePresentationWrapper[];
```
nklomp, line 94: This is not correct. According to https://openid.net/specs/openid-connect-4-verifiable-presentations-1_0.html#name-siop-with-vp_token we should have a vp_token here

nklomp: The above explicitly is about how it should look like in SIOP for both the request and response

nklomp (changes requested):
The vp matching code is much nicer now. There is however something off with the payload location according to the spec. It explicitly mentions where to put it in the SIOP case. We also need to double check where we put the definition on the RP side. As it is also specific about that and we are probably not following ithttps://openid.net/specs/openid-connect-4-verifiable-presentations-1_0.html#name-siop-with-vp_token

PR #6, opened 2021-10-06
PEX changes/fixes. Need test fixes

nklomp:
WIP/Draft PR. Allow multiple locations for VPs and claims according to spec. Move logic from RP/VP to Request/Response.
Remove payload usage from opts. Fix logic errors.

Multiple tests are still failing


src/main/types/SIOP.types.ts
```diff
@@ -90,15 +90,67 @@ export interface AuthenticationResponsePayload extends JWTPayload {
...
+export interface IdTokenClaimPayload {
+  verifiable_presentations?: PresentationDefinition[];
+  [x: string]: unknown;
+}
+export interface VpTokenClaimPayload {
+  verifiable_presentation?: PresentationDefinition;
+  [x: string]: unknown;
+}
+
+export interface ClaimOpts {
```
sksadjad, line 119: based on [embedded-verifiable-present](https://openid.net/specs/openid-connect-4-verifiable-presentations-1_0.html#name-embedded-verifiable-present) and [name-vp-token](https://openid.net/specs/openid-connect-4-verifiable-presentations-1_0.html#name-vp-token) this can't be valid

src/main/types/SIOP.types.ts
```diff
@@ -90,15 +90,67 @@ export interface AuthenticationResponsePayload extends JWTPayload {
...
+export interface OidcClaimRequest {
+  [x: string]: null | OidcClaimJson;
+}*/
+
+export interface IdTokenClaimPayload {
+  verifiable_presentations?: PresentationDefinition[];
+  [x: string]: unknown;
+}
+export interface VpTokenClaimPayload {
+  verifiable_presentation?: PresentationDefinition;
```
sksadjad, line 115: based on [name-embedded-verifiable-present](https://openid.net/specs/openid-connect-4-verifiable-presentations-1_0.html#name-embedded-verifiable-present) the name should be `verifiable_presentations` and in it we will have a _single_ `presentation_definition` object

src/main/types/SIOP.types.ts
```diff
@@ -50,7 +50,7 @@ export interface RequestRegistrationPayload {
 
 export interface VerifiedAuthenticationRequestWithJWT extends VerifiedJWT {
   payload: AuthenticationRequestPayload; // The unsigned Authentication Request payload
-  presentationDefinition?: PresentationDefinition; // The optional presentation definition object that the RP requests
+  presentationDefinitions?: PresentationDefinitionWithLocation[]; // The optional presentation definition objects that the RP requests
```
sksadjad, line 53: this should be single

src/main/types/SIOP.types.ts
```diff
@@ -90,15 +90,67 @@ export interface AuthenticationResponsePayload extends JWTPayload {
...
+  [x: string]: unknown;
+}
+export interface DIDDocument extends DIFDIDDocument {
+  owner?: string;
+  created?: string;
+  updated?: string;
+  proof?: LinkedDataProof;
+}
+
+export interface PresentationDefinitionWithLocation {
```
sksadjad, line 135: based on [section-11.6.1.1](https://openid.net/specs/openid-connect-4-verifiable-presentations-1_0.html#section-11.6.1.1) the definition (which should be `presentation_definition`) have to be inside the `vp_token` object and therefore `PresentationLocation` can't be an enum

test/AuthenticationRequest.request.spec.ts
```diff
@@ -381,24 +381,24 @@ describe('create Request JWT should', () => {
...
-              {
-                'id': 'Ontario Health Insurance Plan',
-                'schema': [
-                  {
-                    'uri': 'https://did.itsourweb.org:3000/smartcredential/Ontario-Health-Insurance-Plan'
-                  }
-                ]
-              }
-            ]
+        presentationDefinitions: [
```
sksadjad, line 384: this is why I didn't change the actual code and just wrote review, in the past two hours I'm trying to find examples about the spec that have this kind of format, but I couldn't find any

sksadjad (changes requested):
lots of confusion on my part, I do not understand the meaning behind the recent changes.
on a minor note, looks like your IDE messed up our styling (in the imports for example)

src/main/types/SIOP.types.ts
nklomp: These are options the user of the library provides. These are not payloads. It gets translated into a payload later.

I mentioned the other day to you that you should not pass in payloads as a user because it exposes too many the internals and you loose strength (like in this case having a single object that can end up in multiple forms in the eventual payload)

src/main/types/SIOP.types.ts
nklomp: It depends on the location, but indeed I probably haven't included both the singular (vp_token) https://openid.net/specs/openid-connect-4-verifiable-presentations-1_0.html#name-siop-with-vp_token) and plural (id_token) case for the payload

src/main/types/SIOP.types.ts
nklomp: No because this is an interpreted field not a payload

test/AuthenticationRequest.request.spec.ts
nklomp: This is an example of using it in a vp_token: see https://openid.net/specs/openid-connect-4-verifiable-presentations-1_0.html#section-11.6.1.1

PR #11, opened 2022-02-09
MYC-146 added SuppliedSignature functionality

BtencateSphereon:


src/main/OPBuilder.ts
```diff
@@ -1,14 +1,23 @@
 import { getUniResolver } from '@sphereon/did-uni-client';
+import {EcdsaSignature} from 'did-jwt/lib/util';
```
nklomp, line 2: Formatting

src/main/types/JWT.types.ts
```diff
@@ -1,7 +1,8 @@
+import { EcdsaSignature } from 'did-jwt/lib/util';
 import type { DIDResolutionResult, VerificationMethod } from 'did-resolver';
 // import type {JWK} from "jose/types";
 
-export type Signer = (data: string | Uint8Array) => Promise<string>;
+export type Signer = (data: string | Uint8Array) => Promise<EcdsaSignature | string>;
```
nklomp, line 5: Guess we should add a comment that this is the signer type/interface conforming to the DID-JWT module

nklomp (approved):
Still the packages/yarn issue. Rest looks okay though

edit: It seems you need to run prettier

BtencateSphereon:
test are failing, looking at it now

PR #15, opened 2022-08-04
WIP: first batch of suggestions

sksadjad:
this can't be merged in it's current form. it's created just to start the conversation about our next set of changes


sksadjad:
@nklomp 
looked for these fields in these docs:
https://openid.net/specs/openid-connect-4-verifiable-presentations-1_0-00.html (20 May 2021)
https://openid.net/specs/openid-connect-4-verifiable-presentations-1_0-04.html (9 September 2021)
https://openid.net/specs/openid-connect-4-verifiable-presentations-1_0-06.html (10 December 2021)
https://openid.net/specs/openid-connect-4-verifiable-presentations-1_0-07.html (17 December 2021)
https://openid.net/specs/openid-connect-4-verifiable-presentations-1_0-08.html (28 January 2022)
https://openid.net/specs/openid-connect-4-verifiable-presentations-1_0-10.html (22 April 2022)
https://openid.net/specs/openid-4-verifiable-presentations-1_0.html (1_0-12 - 21 June 2022)
and couldn't find the values that we were talking about. as you can see from the version `1_0-00` to `1_0-12` there are some revisions missing, I couldn't find them anywhere to compare

src/main/AuthenticationRequest.ts
```diff
@@ -192,7 +192,11 @@ function createClaimsPayload(opts: SIOP.ClaimOpts): ClaimPayload {
           // There can only be one definition in the vp_token according to the spec
           throw new Error(SIOPErrors.REQUEST_CLAIMS_PRESENTATION_DEFINITION_NOT_VALID);
         } else {
-          vp_token = { presentation_definition: def.definition };
+          vp_token = {
+            nonce: 'NONCE_STRING',
```
nklomp, line 196: Hmzz, there needs to be a TODO here, since the nonce is not static of course

src/main/AuthenticationRequestRegistration.ts
```diff
@@ -32,10 +32,16 @@ export function createRequestRegistration(opts: SIOP.RequestRegistrationOpts): {
...
-    did_methods_supported: opts.didMethodsSupported || ['did:eosio:', 'did:ethr:', 'did:factom:', 'did:lto:'],
-    subject_identifiers_supported: opts.subjectIdentifiersSupported || SIOP.SubjectIdentifierType.DID,
-    credential_formats_supported: opts.credentialFormatsSupported || [],
+    authorization_endpoint: opts.authorizationEndpoint,
+    id_token_signing_alg_values_supported: opts.idTokenSigningAlgValuesSupported,
+    request_object_signing_alg_values_supported: opts.requestObjectSigningAlgValuesSupported,
+    response_types_supported: opts.responseTypesSupported,
+    scopes_supported: opts.scopesSupported,
+    subject_types_supported: opts.subjectTypesSupported,
+    subject_syntax_types_supported: opts.subjectSyntaxTypesSupported || ['did:eosio:', 'did:ethr:', 'did:factom:', 'did:lto:'],
```
nklomp, line 44: The did:eosio can be removed. And added: did:key and did:ion, so we can conform to the new JWT VC Presentation Profile

nklomp: We also need to create a Ticket/TODO to start supporting JWK thumbprints, which is easy enough to create: https://www.rfc-editor.org/rfc/rfc7638

src/main/OP.ts
```diff
@@ -168,8 +175,10 @@ function createVerifyRequestOptsFromBuilderOrExistingOpts(opts: { builder?: OPBu
         verification: {
           mode: VerificationMode.INTERNAL,
           resolveOpts: {
-            didMethods: opts.builder.didMethods,
-            resolver: opts.builder.resolver ? getResolver({ resolver: opts.builder.resolver }) : getResolver({ didMethods: opts.builder.didMethods }),
+            didMethods: opts.builder.subjectSyntaxTypesSupported,
```
nklomp, line 178: We need to do filtering on did: in here, and add a todo for support of other subjectSyntaxTypes

src/main/OPBuilder.ts
```diff
@@ -1,33 +1,51 @@
...
   responseMode?: ResponseMode;
+  responseRegistration: ResponseRegistrationOpts;
+  responseTypesSupported: ResponseType[] | ResponseType;
   // did: string;
   // vp?: VerifiablePresentation;
-  expiresIn?: number;
   resolver?: Resolvable;
+  signatureType: InternalSignature | ExternalSignature | SuppliedSignature;
+  subjectSyntaxTypesSupported: string[] = [];
+  vpFormats: Format;
```
nklomp, line 34: Is vpFormats a required argument?

src/main/OPBuilder.ts
```diff
@@ -1,33 +1,51 @@
...
+  addResponseTypesSupported(responseTypesSupported: ResponseType[] | ResponseType): OPBuilder {
+    this.responseTypesSupported = responseTypesSupported;
+    return this;
+  }
 
-  addCredentialFormat(credentialFormat: CredentialFormat): OPBuilder {
-    this.credentialFormats.push(credentialFormat);
+  addIssuer(issuer: ResponseIss): OPBuilder {
+    this.issuer = issuer;
     return this;
```
nklomp, line 49: I am missing some builder methods for the new properties

src/main/OPBuilder.ts
```diff
@@ -62,8 +80,13 @@ export default class OPBuilder {
...
   }
 
+  //TODO: fill up with the actual values
   registrationBy(registrationBy: PassBy, refUri?: string): OPBuilder {
     this.responseRegistration = {
+      vpFormats: this.vpFormats,
+      issuer: undefined,
```
nklomp, line 87: This doesn't look right (the undefined and empty arrays. Needs to have a clear TODO

src/main/RP.ts
```diff
@@ -119,8 +124,10 @@ function createVerifyResponseOptsFromBuilderOrExistingOpts(opts: { builder?: RPB
         verification: {
           mode: VerificationMode.INTERNAL,
           resolveOpts: {
-            didMethods: opts.builder.didMethods,
-            resolver: opts.builder.resolver ? getResolver({ resolver: opts.builder.resolver }) : getResolver({ didMethods: opts.builder.didMethods }),
+            didMethods: opts.builder.subjectSyntaxTypesSupported,
```
nklomp, line 127: Needs a did: filter in there, as subjectTypesSupported is bigger than DIDs alone. Also needs a TODO to handle other types like JWK-thumbprint

src/main/RPBuilder.ts
```diff
@@ -35,9 +38,21 @@ export default class RPBuilder {
   claims?: ClaimOpts;
 
   // claims?: ClaimPayload;
+  vpFormatsSupported: Format;
+  idTokenSigningAlgValuesSupported: SigningAlgo[];
```
nklomp, line 42: Are all of these required parameters properties?

src/main/functions/DidSiopMetadata.ts
```diff
@@ -1,18 +1,20 @@
...
-  const credentials = supportedCredentialsFormats(rpMetadata.credential_formats_supported, opMetadata.credential_formats_supported);
-  const isDid = verifySubjectIdentifiers(rpMetadata.subject_identifiers_supported);
-  if (isDid && rpMetadata.did_methods_supported) {
-    methods = supportedDidMethods(rpMetadata.did_methods_supported, opMetadata.did_methods_supported);
-  } else if (isDid && (!rpMetadata.did_methods_supported || !rpMetadata.did_methods_supported.length)) {
-    if (opMetadata.did_methods_supported || opMetadata.did_methods_supported.length) {
-      methods = [...opMetadata.did_methods_supported];
+  let subjectSyntaxTypesSupported = [];
+  const credentials = supportedCredentialsFormats(rpMetadata.vp_formats, opMetadata.vp_formats);
+  const isDid = verifySubjectIdentifiers(rpMetadata.subject_syntax_types_supported);
```
nklomp, line 9: Needs a TODO, since syntax_types_Supported can contain non DIDs

src/main/functions/DidSiopMetadata.ts
```diff
@@ -1,18 +1,20 @@
...
-  let methods = [];
-  const credentials = supportedCredentialsFormats(rpMetadata.credential_formats_supported, opMetadata.credential_formats_supported);
-  const isDid = verifySubjectIdentifiers(rpMetadata.subject_identifiers_supported);
-  if (isDid && rpMetadata.did_methods_supported) {
-    methods = supportedDidMethods(rpMetadata.did_methods_supported, opMetadata.did_methods_supported);
-  } else if (isDid && (!rpMetadata.did_methods_supported || !rpMetadata.did_methods_supported.length)) {
-    if (opMetadata.did_methods_supported || opMetadata.did_methods_supported.length) {
-      methods = [...opMetadata.did_methods_supported];
+  let subjectSyntaxTypesSupported = [];
+  const credentials = supportedCredentialsFormats(rpMetadata.vp_formats, opMetadata.vp_formats);
```
nklomp, line 8: Was also in the original code, but we need to rename credentials with credentialFormats, as the name is highly confusing

src/main/functions/DidSiopMetadata.ts
```diff
@@ -30,29 +32,50 @@ function getIntersection<T>(rpMetadata: Array<T> | T, opMetadata: Array<T> | T):
   return arrayA.filter((value) => arrayB.includes(value));
 }
 
-function verifySubjectIdentifiers(subjectIdentifiers: SubjectIdentifierType | SubjectIdentifierType[]): boolean {
-  if (subjectIdentifiers || subjectIdentifiers.length) {
-    if (Array.isArray(subjectIdentifiers)) {
-      return subjectIdentifiers.includes(SIOP.SubjectIdentifierType.DID);
+function verifySubjectIdentifiers(subjectSyntaxTypesSupported: string[]): boolean {
```
nklomp, line 35: Are you really sure that a single non array element isn't allowed anymore?

src/main/functions/DidSiopMetadata.ts
```diff
@@ -30,29 +32,50 @@ function getIntersection<T>(rpMetadata: Array<T> | T, opMetadata: Array<T> | T):
...
+  if (!supportedDidMethods.length || (supportedDidMethods.length === 1 && supportedDidMethods[0] === SubjectIdentifierType.DID)) {
     throw Error(SIOPErrors.DID_METHODS_NOT_SUPORTED);
   }
   return supportedDidMethods;
 }
 
-function supportedCredentialsFormats(rpCredentials: string[] | string, opCredentials: string[] | string): Array<string> {
-  const supportedCredentials = getIntersection(rpCredentials, opCredentials);
+export function supportedCredentialsFormats(rpCredentials: Format, opCredentials: Format): Format {
+  if (!rpCredentials || !opCredentials || !Object.keys(rpCredentials).length || !Object.keys(opCredentials).length) {
```
nklomp, line 53: Hmzz, not sure if this is correct. If the rp doesn't provide formats (we need to rename rpCredentials and opCredentials BTW), they are signaling they support everything right? Or at least that you probably will only get an error later down the line. Similar for the OP. If they do not provide the info I would assume they are signaling to support all kinds of formats.

src/main/functions/DidSiopMetadata.ts
```diff
@@ -30,29 +32,50 @@ function getIntersection<T>(rpMetadata: Array<T> | T, opMetadata: Array<T> | T):
...
+    Object.keys(rpCredentials[crFormat]).forEach((k) => (methodKeyRP = k));
+    Object.keys(opCredentials[crFormat]).forEach((k) => (methodKeyOP = k));
+    if (methodKeyRP !== methodKeyOP) {
+      throw new Error(SIOPErrors.CREDENTIAL_FORMATS_NOT_SUPPORTED);
+    }
+    const algs = getIntersection(rpAlgs, opAlgs);
+    if (!algs.length) {
+      throw new Error(SIOPErrors.CREDENTIAL_FORMATS_NOT_SUPPORTED);
+    }
+    intersectionFormat[crFormat] = {};
```
nklomp, line 77: Wouldn't it make sense to have the intersection method do most of this and have it return the more complex object?

src/main/types/SIOP.types.ts
```diff
@@ -10,38 +10,48 @@ export const expirationTime = 10 * 60;
...
   scope: string;
   response_type: ResponseType;
   client_id: string; // did of RP
   redirect_uri: string;
   id_token_hint?: string; // TODO:  idtokenhint parameter value, as specified in Section 3.1.2. If the ID Token is encrypted to the Self-Issued OP, the sub (subject) of the signed ID Token MUST be sent as the kid (Key ID) of the JWE.
   // iss: string;
-  response_mode: ResponseMode;
-  response_context: ResponseContext;
-
+  //response_mode: ResponseMode;
```
nklomp, line 39: Hmzz, I believe Response_mode wat a lib implementation detail, not sure if it can be removed right now

test/AuthenticationRequest.request.spec.ts
```diff
@@ -110,9 +128,17 @@ describe('create Request Uri should', () => {
...
       },
       registration: {
-        didMethodsSupported: ['did:ethr:'],
-        subjectIdentifiersSupported: SubjectIdentifierType.DID,
-        credentialFormatsSupported: CredentialFormat.JSON_LD,
+        idTokenSigningAlgValuesSupported: [SigningAlgo.EDDSA, SigningAlgo.ES256],
+        requestObjectSigningAlgValuesSupported: [SigningAlgo.EDDSA, SigningAlgo.ES256],
+        responseTypesSupported: [ResponseType.ID_TOKEN],
+        scopesSupported: [Scope.OPENID_DIDAUTHN, Scope.OPENID],
+        subjectSyntaxTypesSupported: ['did:ethr:', SubjectIdentifierType.DID],
```
nklomp, line 135: did:ethr comes from nowhere ;)

nklomp (changes requested):
Did an initial pass in terms of review.

src/main/AuthenticationRequestRegistration.ts
sksadjad: removed `did:eosio`, added: `did:key` and `did:ion`

src/main/OP.ts
sksadjad: added filter and a ticket https://sphereon.atlassian.net/browse/VDX-126

src/main/OPBuilder.ts
sksadjad: yes it is https://openid.net/specs/openid-4-verifiable-presentations-1_0.html#section-8.2.3.1

src/main/AuthenticationRequest.ts
sksadjad: added

src/main/OPBuilder.ts
sksadjad: added

src/main/OPBuilder.ts
sksadjad: fixed

src/main/RP.ts
sksadjad: added

src/main/RPBuilder.ts
sksadjad: yes and made a comment about required ones in the `SIOP.types`

src/main/functions/DidSiopMetadata.ts
sksadjad: added TODO for VDX-126

src/main/functions/DidSiopMetadata.ts
sksadjad: yes, I haven't seen singular form anywhere

src/main/functions/DidSiopMetadata.ts
sksadjad: this is a little bit tricky, `vp_format` is a required object

src/main/functions/DidSiopMetadata.ts
nklomp: If you haven't seen it, is there a normative text in the spec that makes clear it only can be an array. Otherwise we are now removing a more relaxed handling of the object. In other words, if we support both it will not hurt anyone.

src/main/functions/DidSiopMetadata.ts
sksadjad: changed a little bit, but the point is the intersection method is generic and I don't want to mess up the other methods using it, so I created a wrapper for it

src/main/OP.ts
```diff
@@ -149,11 +149,46 @@ function createResponseOptsFromBuilderOrExistingOpts(opts: { builder?: OPBuilder
...
+          jwksUri: opts.builder.responseRegistration.jwksUri,
+          registrationEndpoint: opts.builder.responseRegistration.registrationEndpoint,
+          scopesSupported: opts.builder.responseRegistration.scopesSupported,
+          responseTypesSupported: opts.builder.responseRegistration.responseTypesSupported,
+          responseModesSupported: opts.builder.responseRegistration.responseModesSupported,
+          grantTypesSupported: opts.builder.responseRegistration.grantTypesSupported,
+          acrValuesSupported: opts.builder.responseRegistration.acrValuesSupported,
+          subjectTypesSupported: opts.builder.responseRegistration.subjectTypesSupported,
+          idTokenSigningAlgValuesSupported: opts.builder.responseRegistration.idTokenSigningAlgValuesSupported,
+          idTokenEncryptionAlgValuesSupported: opts.builder.responseRegistration.idTokenEncryptionAlgValuesSupported,
```
nklomp, line 164: Hmzz, we probably want to do some refactoring and destructuring in here.

src/main/types/SIOP.types.ts
sksadjad: added back

test/AuthenticationRequest.request.spec.ts
sksadjad: no, it was in the previous version with a little change:
```js
didMethodsSupported: ['did:ethr:'],
subjectIdentifiersSupported: SubjectIdentifierType.DID,
```

src/main/functions/DidSiopMetadata.ts
sksadjad: `vp_formats: REQUIRED. An object defining the formats, proof types and algorithms of verifiable presentations and verifiable credentials that a RP supports. Valid values are defined in the table in Section 6.7.3. of [[OpenID.VCI](https://openid.net/specs/openid-4-verifiable-presentations-1_0.html#OpenID.VCI)] and include jwt_vc, ldp_vc, jwt_jp and ldp_vp. Formats identifiers not in the table may be supported.`

sksadjad: so I don't think it can be empty

src/main/functions/DidSiopMetadata.ts
sksadjad: I'm not sure what you mean, but both [jwt-vc-presentation-profile](https://identity.foundation/jwt-vc-presentation-profile/#verifierrp-registration-metadata) and [SIOPv2](https://openid.net/specs/openid-connect-self-issued-v2-1_0-06.html#section-9.2-5.8.1) are only talking about jsonarrays

src/main/OP.ts
sksadjad: yeah right now it's quite messy

sksadjad:
Hey @nklomp if you agree with the content, I can start to make it a nicer pr, right now, I'm just trying to make it work, and I'm passing all the previous tests

sksadjad:
Hey @nklomp if you agree with the content, I can start to make it a nicer pr, right now, I'm just trying to make it work, and I'm passing all the previous tests

src/main/functions/DidSiopMetadata.ts
nklomp: Agree, it needs to be explicit

test/AuthenticationRequest.request.spec.ts
nklomp: Yeah it was there for demo purposes

src/main/AuthenticationRequestRegistration.ts
nklomp: Sory I meant did:web instead of did:key. Maybe we should also just remove did:ethr. did:factom and did:lto, to make it less opiniated for others. So if not supplied we only support did-web and did-ion conforming to the Interop profile

src/main/functions/DidSiopMetadata.ts
nklomp: The first spec is only something on top of the underlying specs, and should never be taken into account in a lib which can also be used outside of the interop profile. 

The second does make it clear that it needs to be an array. But since there is no schema yet, I would rather have us be relaxed about it. So whenever the arguments is being set from a developer we expect an array to be passed in. But of it comes in as a payload we can accept a single string instead of an array.

nklomp:
Some small remarks left

src/main/functions/DidSiopMetadata.ts
sksadjad: changed all of the list types to accept a singular value as well

src/main/AuthenticationRequestRegistration.ts
sksadjad: changed to: `subject_syntax_types_supported: opts.subjectSyntaxTypesSupported || ['did:web:', 'did:ion:']`

nklomp:
One comment lef, which is simply removing the did:ethr from that exact location as it was demo code

test/AuthenticationRequest.request.spec.ts
nklomp: Is a test class so probably there for a reason

nklomp (approved):

PR #19, opened 2022-08-22
Feature/vdx 120 presentation definition

sksadjad:
adding the pd ref ability
- in the previous version we could only send the value of a PD, this pr enables us to send a `presentation_definition_uri` instead. Note that you can't send both `presentation_definition` and `presentation_definition_uri` at the same time

### DIFF
--- a/src/main/PresentationExchange.ts
+++ b/src/main/PresentationExchange.ts
@@ -51,8 +51,6 @@ export class PresentationExchange {
     }
 
     function sign(params: PresentationSignCallBackParams): Promise<IVerifiablePresentation> {
-      // console.log('##### SIGN CALLBACK IMPLEMENTATION NEEDED FOR VP');
-      // console.log(params);
       return Promise.resolve(params.presentation as IVerifiablePresentation);
     }
 
@@ -191,7 +189,6 @@ export class PresentationExchange {
         const pd: PresentationDefinitionV1 | PresentationDefinitionV2 = (await getWithUrl(definitionRef[0].value)) as unknown as
           | PresentationDefinitionV1
           | PresentationDefinitionV2;
-        console.log(JSON.stringify(pd));
         allDefinitions = handleAddingSinglePDInIdToken(pd, allDefinitions);
       }
     }
--- a/src/main/PresentationExchange.ts
+++ b/src/main/PresentationExchange.ts
@@ -138,58 +138,55 @@ export class PresentationExchange {
    * @param obj: object that can have a presentation_definition inside
    */
   public static async findValidPresentationDefinitions(obj: JWTPayload): Promise<PresentationDefinitionWithLocation[]> {
-    let allDefinitions: PresentationDefinitionWithLocation[];
+    const allDefinitions: PresentationDefinitionWithLocation[] = [];
+
     async function extractPDFromVPToken() {
       const vpTokens = extractDataFromPath(obj, '$..vp_token.presentation_definition');
       const vpTokenRef = extractDataFromPath(obj, '$..vp_token.presentation_definition_uri');
       if (vpTokens && vpTokens.length && vpTokenRef && vpTokenRef.length) {
-        throw new Error(SIOPErrors.REQUEST_CLAIMS_CANT_SEND_PRESENTATION_DEFINITION_BY_RE_AND_VAL);
+        throw new Error(SIOPErrors.REQUEST_CLAIMS_PRESENTATION_DEFINITION_BY_REF_AND_VALUE_NON_EXCLUSIVE);
       }
-      if (vpTokens && vpTokens.length == 1) {
-        PresentationExchange.assertValidPresentationDefinition(vpTokens[0].value);
-        allDefinitions = [{ definition: vpTokens[0].value, location: PresentationLocation.VP_TOKEN }];
-      }
-      if (vpTokens && vpTokens.length > 1) {
-        throw new Error(SIOPErrors.REQUEST_CLAIMS_PRESENTATION_DEFINITION_NOT_VALID);
-      }
-      if (vpTokenRef && vpTokenRef.length) {
-        const result = await getWithUrl(vpTokenRef[0].value);
-        const pd: PresentationDefinitionV1 | PresentationDefinitionV2 = await result.json();
-        PresentationExchange.assertValidPresentationDefinition(pd);
-        allDefinitions = [{ definition: pd, location: PresentationLocation.VP_TOKEN }];
+      if (vpTokens && vpTokens.length) {
+        vpTokens.forEach((vpToken) => {
+          PresentationExchange.assertValidPresentationDefinition(vpToken.value);
+          allDefinitions.push({ definition: vpToken.value, location: PresentationLocation.VP_TOKEN });
+        });
+      } else if (vpTokenRef && vpTokenRef.length) {
+        for (const vptr of vpTokenRef) {
+          const pd: PresentationDefinitionV1 | PresentationDefinitionV2 = (await getWithUrl(vptr.value)) as unknown as
+            | PresentationDefinitionV1
+            | PresentationDefinitionV2;
+          PresentationExchange.assertValidPresentationDefinition(pd);
+          allDefinitions.push({ definition: pd, location: PresentationLocation.VP_TOKEN });
+        }
       }
     }
 
-    function handleAddingSinglePDInIdToken(definition, allDefinitions: PresentationDefinitionWithLocation[]): PresentationDefinitionWithLocation[] {
+    function addSingleIdTokenPDToPDs(definition) {
       PresentationExchange.assertValidPresentationDefinition(definition.value);
-      if (definition.path.includes(PresentationLocation.ID_TOKEN)) {
-        const defWithLocation = { definition: definition.value, location: PresentationLocation.ID_TOKEN };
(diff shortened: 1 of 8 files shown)


src/main/PresentationExchange.ts
```diff
@@ -138,47 +137,70 @@ export class PresentationExchange {
...
     let allDefinitions: PresentationDefinitionWithLocation[];
-
-    function extractPDFromVPToken() {
+    async function extractPDFromVPToken() {
       const vpTokens = extractDataFromPath(obj, '$..vp_token.presentation_definition');
+      const vpTokenRef = extractDataFromPath(obj, '$..vp_token.presentation_definition_uri');
+      if (vpTokens && vpTokens.length && vpTokenRef && vpTokenRef.length) {
+        throw new Error(SIOPErrors.REQUEST_CLAIMS_CANT_SEND_PRESENTATION_DEFINITION_BY_RE_AND_VAL);
+      }
+      if (vpTokens && vpTokens.length == 1) {
```
nklomp, line 148: To make all of this a bit more readable suggest to change this to 

```
if (.....length ==1) {
...
} else if (....length > 1) {
....
} else if (ref && ref.length) {
...
}
```

src/main/PresentationExchange.ts
```diff
@@ -138,47 +137,70 @@ export class PresentationExchange {
...
+      PresentationExchange.assertValidPresentationDefinition(definition.value);
+      if (definition.path.includes(PresentationLocation.ID_TOKEN)) {
+        const defWithLocation = { definition: definition.value, location: PresentationLocation.ID_TOKEN };
+        if (!allDefinitions) {
+          allDefinitions = [defWithLocation];
+        } else {
+          allDefinitions.push(defWithLocation);
+        }
+      } else {
+        throw new Error(SIOPErrors.REQUEST_CLAIMS_PRESENTATION_DEFINITION_NOT_VALID);
```
nklomp, line 173: Little bit easier to read if the expression is negated and this error is thrown after the expression. Then the rest can continue in the function body instead if in the if body

src/main/PresentationExchange.ts
```diff
@@ -138,47 +137,70 @@ export class PresentationExchange {
...
+        allDefinitions = [{ definition: pd, location: PresentationLocation.VP_TOKEN }];
+      }
+    }
 
-      if (vpTokens) {
-        if (vpTokens.length == 1) {
-          PresentationExchange.assertValidPresentationDefinition(vpTokens[0].value);
-          allDefinitions = [{ definition: vpTokens[0].value, location: PresentationLocation.VP_TOKEN }];
-        } else if (vpTokens.length > 1) throw new Error(SIOPErrors.REQUEST_CLAIMS_PRESENTATION_DEFINITION_NOT_VALID);
+    function handleAddingSinglePDInIdToken(definition, allDefinitions: PresentationDefinitionWithLocation[]): PresentationDefinitionWithLocation[] {
```
nklomp, line 163: addSingleIdTokenPDToPDs to make it a bit more clear that the definition is being added. Actually to make it more clear the input array is modified this probably should return a void

src/main/PresentationExchange.ts
```diff
@@ -138,47 +137,70 @@ export class PresentationExchange {
...
+      PresentationExchange.assertValidPresentationDefinition(definition.value);
+      if (definition.path.includes(PresentationLocation.ID_TOKEN)) {
+        const defWithLocation = { definition: definition.value, location: PresentationLocation.ID_TOKEN };
+        if (!allDefinitions) {
+          allDefinitions = [defWithLocation];
```
nklomp, line 168: If we init the allDefinition as an empty array we can always just push. It also is probably less error prone for the `allDefinitions = [{ definition: pd, location: PresentationLocation.VP_TOKEN }];` type of constructs elsewhere

src/main/functions/HttpUtils.ts
```diff
@@ -43,3 +43,17 @@ export async function postAuthenticationResponseJwt(url: string, jwt: string): P
...
+  try {
+    const response = await fetch(url, {
+      method: 'GET',
+    });
+    if (!response || !response.status || (response.status !== 200 && response.status !== 201)) {
+      throw new Error(`${SIOPErrors.RESPONSE_STATUS_UNEXPECTED} ${response.status}:${response.statusText}, ${await response.json()}`);
+    }
+    return response;
+  } catch (error) {
+    throw new Error(`${(error as Error).message}`);
```
nklomp, line 57: Hmzz, cant the catch be moved up. This is also catching the error being created in case the response status is non 200.

src/main/functions/HttpUtils.ts
```diff
@@ -43,3 +43,17 @@ export async function postAuthenticationResponseJwt(url: string, jwt: string): P
     throw new Error(`${(error as Error).message}`);
   }
 }
+
+export async function getWithUrl(url: string): Promise<Response> {
+  try {
+    const response = await fetch(url, {
+      method: 'GET',
+    });
+    if (!response || !response.status || (response.status !== 200 && response.status !== 201)) {
```
nklomp, line 52: &gt;=400 instead of 200 && 201?

src/main/types/Errors.ts
```diff
@@ -36,6 +36,7 @@ enum SIOPErrors {
   NO_SELFISSUED_ISS = 'The Response Token Issuer Claim (iss) MUST be https://self-isued.me',
   NO_SUB_TYPE = 'No or empty sub_type found in JWT payload',
   REGISTRATION_NOT_SET = 'Registration metadata not set.',
+  REQUEST_CLAIMS_CANT_SEND_PRESENTATION_DEFINITION_BY_RE_AND_VAL = "Request claims can't have both 'presentation_definition' and 'presentation_definition_uri'",
```
nklomp, line 39: RE needs to be REF. Probably something like REQUEST_CLAIMS_PRESENTATION_DEFINITION_BY_REF_AND_VALUE_NON_EXCLUSIVE

test/PresentationExchange.spec.ts
```diff
@@ -95,6 +97,50 @@ async function getPayload(): Promise<AuthenticationRequestPayload> {
   };
 }
 
+async function getPayload_PD_Ref(): Promise<AuthenticationRequestPayload> {
```
nklomp, line 100: Why are we all of a sudden using a different naming convention (camelcase/snakecase)?

test/PresentationExchange.spec.ts
```diff
@@ -123,9 +169,68 @@ function getVCs(): IVerifiableCredential[] {
...
+      verifyOpts: null,
+    };
+    try {
+      await PresentationExchange.validatePresentationAgainstDefinition(verifiedJwt.presentationDefinitions[0].definition, {
+        '@context': ['https://www.w3.org/2018/credentials/v1'],
+        type: ['VerifiablePresentation'],
+        verifiableCredential: vcs,
+      });
+    } catch (e) {
+      expect(e.message).toContain(SIOPErrors.COULD_NOT_FIND_VCS_MATCHING_PD);
```
nklomp, line 193: You can also put the expect on the validate function and then use .rejects instead of catching the Error and then doin an expect btw.

nklomp (changes requested):
See remarks

test/PresentationExchange.spec.ts
sksadjad: changed

src/main/types/Errors.ts
sksadjad: changed

src/main/functions/HttpUtils.ts
sksadjad: changed

src/main/PresentationExchange.ts
sksadjad: changed

src/main/PresentationExchange.ts
sksadjad: changed

src/main/PresentationExchange.ts
sksadjad: changed

src/main/PresentationExchange.ts
sksadjad: changed

src/main/PresentationExchange.ts
```diff
@@ -138,47 +137,70 @@ export class PresentationExchange {
...
+      const vpTokenRef = extractDataFromPath(obj, '$..vp_token.presentation_definition_uri');
+      if (vpTokens && vpTokens.length && vpTokenRef && vpTokenRef.length) {
+        throw new Error(SIOPErrors.REQUEST_CLAIMS_CANT_SEND_PRESENTATION_DEFINITION_BY_RE_AND_VAL);
+      }
+      if (vpTokens && vpTokens.length == 1) {
+        PresentationExchange.assertValidPresentationDefinition(vpTokens[0].value);
+        allDefinitions = [{ definition: vpTokens[0].value, location: PresentationLocation.VP_TOKEN }];
+      }
+      if (vpTokens && vpTokens.length > 1) {
+        throw new Error(SIOPErrors.REQUEST_CLAIMS_PRESENTATION_DEFINITION_NOT_VALID);
```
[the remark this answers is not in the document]
sksadjad: changed

nklomp (approved):

PR #22, opened 2022-08-26
WIP: Feature/vdx 103 linked domain validation

sksadjad:
verifying linked domain credentials


src/main/RP.ts
```diff
@@ -1,18 +1,22 @@
+import { DomainLinkageCredential, IVerifyCredentialResult, WellKnownDidVerifier } from '@sphereon/wellknown-dids-client';
+import WDCErrors from '@sphereon/wellknown-dids-client/dist/constants/Errors';
```
nklomp, line 2: Hmzz, apparently this isn't being exported properly. We need to fix that.

src/main/functions/DIDResolution.ts
```diff
@@ -23,7 +24,3 @@ export function getResolver(opts: ResolveOpts): Resolvable {
   }
   return new Resolver(...uniResolvers);
 }
-
-export async function resolveDidDocument(did: string, opts?: ResolveOpts): Promise<DIDDocument> {
```
nklomp, line 27: Is there a specific reason this is removed. Seem pretty useful to me

src/main/RP.ts
```diff
@@ -84,13 +91,50 @@ export class RP {
...
+
+  private async verifyLinkedDomainCredentialVC(vc: DomainLinkageCredential, verifier: WellKnownDidVerifier) {
+    try {
+      await verifier.verifyDomainLinkageCredential({ credential: vc });
+    } catch (e) {
+      if (
+        (e.message != WDCErrors.PROPERTY_LINKED_DIDS_DOES_NOT_CONTAIN_ANY_DOAMIN_LINK_CREDENTIALS &&
+          e.message != WDCErrors.PROPERTY_LINKED_DIDS_NOT_PRESENT &&
+          e.message != WDCErrors.PROPERTY_TYPE_NOT_CONTAIN_VALID_LINKED_DOMAIN) ||
+        this.authRequestOpts.linkedDomainValidationMode != LinkedDomainValidationMode.OPTIONAL
```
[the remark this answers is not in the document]
BtencateSphereon: The library should already give back a response object with possible issues on a did document (services), did configuration resources attached to the origins within a service and the credentials within these resources. I am not sure what the exact issue is but we can discuss this on monday

src/main/AuthenticationRequestRegistration.ts
```diff
@@ -1,38 +1,62 @@
...
 }
 
-export function createRequestRegistrationPayload(opts: SIOP.RequestRegistrationOpts): RequestRegistrationPayload {
+export async function createRequestRegistrationPayload(opts: RequestRegistrationOpts): Promise<RequestRegistrationPayload> {
   assertValidRequestRegistrationOpts(opts);
+
+  const regObj: RPRegistrationMetadataPayload = createRPRegistrationMetadataPayload(opts);
+
+  if (opts.registrationBy.referenceUri) {
+    const regObjToValidate = (await getWithUrl(opts.registrationBy.referenceUri)) as unknown as RPRegistrationMetadataPayload;
```
nklomp, line 33: Is the URL available at this point? We can only perform these types of check during creation at the very end. To ensure consistency

src/main/AuthenticationRequestRegistration.ts
```diff
@@ -1,38 +1,62 @@
...
-  rpRegistrationMetadataPayload: SIOP.RPRegistrationMetadataPayload;
-  opts: SIOP.RequestRegistrationOpts;
-} {
+  rpRegistrationMetadataPayload: RPRegistrationMetadataPayload;
+  opts: RequestRegistrationOpts;
+}> {
+  const requestRegistrationPayload = await createRequestRegistrationPayload(opts);
   return {
-    requestRegistrationPayload: createRequestRegistrationPayload(opts),
+    requestRegistrationPayload: requestRegistrationPayload,
```
nklomp, line 53: Not sure why it was separated, and the next line isn't. If we want to keep that we can have a single requestRegistrationPayload value in the object

src/main/AuthenticationResponse.ts
```diff
@@ -4,19 +4,40 @@ import { JWK } from 'jose/types';
...
+  getNonce,
+  getPublicJWKFromHexPrivateKey,
+  getResolver,
+  getState,
+  getThumbprint,
+  parseJWT,
+  signDidJwtPayload,
+  verifyDidJWT,
+} from './functions';
+import { validateWithDid } from './functions/LinkedDomainValidations';
```
nklomp, line 18: This should be properly exported, so we can import it from ./functions

src/main/AuthenticationResponse.ts
```diff
@@ -90,21 +112,28 @@ export default class AuthenticationResponse {
    *
    * @param jwt ID token to be validated
    * @param verifyOpts
+   * @param linkedDomainValidationMode The policy regarding linkedDomainValidation
    */
-  static async verifyJWT(jwt: string, verifyOpts: VerifyAuthenticationResponseOpts): Promise<VerifiedAuthenticationResponseWithJWT> {
+  static async verifyJWT(
+    jwt: string,
+    verifyOpts: VerifyAuthenticationResponseOpts,
+    linkedDomainValidationMode?: LinkedDomainValidationMode
```
nklomp, line 120: Hmzz, why isn't this part of the actual Verify Options object?

src/main/AuthenticationResponse.ts
```diff
@@ -90,21 +112,28 @@ export default class AuthenticationResponse {
...
     assertValidResponseJWT({ header, payload });
 
-    const verifiedJWT = await verifyDidJWT(jwt, DIDres.getResolver(verifyOpts.verification.resolveOpts), {
+    const verifiedJWT = await verifyDidJWT(jwt, getResolver(verifyOpts.verification.resolveOpts), {
       audience: verifyOpts.audience,
     });
 
-    const issuerDid = DIDJwt.getIssuerDidFromPayload(payload);
+    const issuerDid = getIssuerDidFromPayload(payload);
+    if (linkedDomainValidationMode && linkedDomainValidationMode !== LinkedDomainValidationMode.NEVER)
```
nklomp, line 135: brackets

src/main/AuthenticationRequest.ts
```diff
@@ -66,30 +105,40 @@ export default class AuthenticationRequest {
...
+    AuthenticationRequest.assertValidRegistrationObject(
+      await AuthenticationRequest.getRegistrationObj(verPayload.registration_uri, verPayload.registration)
+    );
+
+    const verifiedJWT = await verifyDidJWT(jwt, getResolver(opts.verification.resolveOpts), options);
     if (!verifiedJWT || !verifiedJWT.payload) {
       throw Error(SIOPErrors.ERROR_VERIFYING_SIGNATURE);
     }
+    if (linkedDomainValidationMode && linkedDomainValidationMode != LinkedDomainValidationMode.NEVER) {
+      validateWithDid(verPayload.iss, linkedDomainValidationMode);
```
nklomp, line 140: Are we really sure the iss field is a DID here?

src/main/OP.ts
```diff
@@ -44,21 +43,26 @@ export class OP {
...
+    requestJwtOrUri: string,
     opts?: { nonce?: string; verification?: InternalVerification | ExternalVerification }
   ): Promise<VerifiedAuthenticationRequestWithJWT> {
-    const jwt = requestJwtorUri.startsWith('ey') ? requestJwtorUri : (await parseAndResolveUri(requestJwtorUri)).jwt;
-    const verifiedJwt = AuthenticationRequest.verifyJWT(jwt, this.newVerifyAuthenticationRequestOpts(opts));
+    const jwt = requestJwtOrUri.startsWith('ey') ? requestJwtOrUri : (await parseAndResolveUri(requestJwtOrUri)).jwt;
+    const verifiedJwt = AuthenticationRequest.verifyJWT(
+      jwt,
+      this.newVerifyAuthenticationRequestOpts(opts),
+      this._authResponseOpts.linkedDomainValidationMode
```
nklomp, line 59: Hmzz, this doesn't look like it is inline with other areas. Probably the opts needs to be extracted into an interface and the domain linkage should be added there

src/main/OP.ts
```diff
@@ -140,6 +145,12 @@ export class OP {
 async function parseAndResolveUri(encodedUri: string) {
   const requestPayload = AuthenticationRequest.parseURI(encodedUri);
   const jwt = requestPayload.request || (await (await fetch(requestPayload.request_uri)).text());
+
+  AuthenticationRequest.assertValidRequestObject(requestPayload);
+  AuthenticationRequest.assertValidRegistrationObject(
+    await AuthenticationRequest.getRegistrationObj(requestPayload.registration_uri, requestPayload.registration)
```
nklomp, line 151: Extract this into an assignment, as you now see what I warned about elsewhere. If this is registration by Ref, we are now hitting the same URI twice in this method (see below).
Actually the line below does the same as the getRegistrationObj method

src/main/RP.ts
```diff
@@ -54,12 +51,12 @@ export class RP {
       claims?: ClaimOpts;
     }
   ): Promise<VerifiedAuthenticationResponseWithJWT> {
-    return AuthenticationResponse.verifyJWT(jwt, this.newVerifyAuthenticationResponseOpts(opts));
+    return AuthenticationResponse.verifyJWT(jwt, this.newVerifyAuthenticationResponseOpts(opts), this._authRequestOpts.linkedDomainValidationMode);
```
nklomp, line 54: This should not become a third param. Should be handled by the newVerifyAuthResponseOpts method

src/main/types/SIOP.types.ts
```diff
@@ -416,6 +416,11 @@ export enum ResponseContext {
   OP = 'op',
 }
 
+export enum LinkedDomainValidationMode {
+  NEVER = 'never', // We don't want to verify Linked domains
+  OPTIONAL = 'optional', // If present, did-auth-siop will check the linked domain
+  ALWAYS = 'always', // We'll always check the linked domains
```
nklomp, line 422: Probably require is a better name. And we should make clear in the comment that this will result in an error if no linked domain is present

nklomp (changes requested):
See remarks

src/main/RP.ts
sksadjad: fixed

src/main/functions/DIDResolution.ts
sksadjad: added back

src/main/functions/DIDResolution.ts
```diff
@@ -13,7 +13,8 @@ export function getResolver(opts: ResolveOpts): Resolvable {
   if (!opts || !opts.subjectSyntaxTypesSupported) {
     throw new Error(SIOPErrors.BAD_PARAMS);
   }
-
+  // 'did:eth:'
```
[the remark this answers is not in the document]
sksadjad: removed

src/main/AuthenticationResponse.ts
sksadjad: changed

src/main/RP.ts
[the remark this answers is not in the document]
nklomp: Discussed. Using another method returns objects. We have a ticket to change this method in the future

nklomp:
PR has conflicts and build is failing

src/main/RPBuilder.ts
```diff
@@ -31,6 +32,7 @@ export default class RPBuilder {
   responseMode?: ResponseMode;
   responseContext?: ResponseContext.RP;
   claims?: ClaimOpts;
+  linkedDomainCheckMode: LinkedDomainValidationMode;
```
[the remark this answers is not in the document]
sksadjad: changed to optional in current version

src/main/AuthenticationRequestRegistration.ts
sksadjad: changed the structure of handling `registration_uri`

src/main/AuthenticationRequestRegistration.ts
sksadjad: changed

src/main/AuthenticationResponse.ts
sksadjad: it is now

src/main/AuthenticationResponse.ts
sksadjad: added

src/main/AuthenticationRequest.ts
sksadjad: yes, we're populating it with a did. of course if we were checking the `registration.issuer` it would be: `ResponseIss.SELF_ISSUED_V2`

src/main/OP.ts
sksadjad: yes, changed

src/main/OP.ts
sksadjad: for this issue, I extracted a method hierarchy call for `getRegistrationObj` and changed it to call it once (per separate method call)

src/main/types/SIOP.types.ts
sksadjad: I'm not sure if you're talking about the name of the enum or enum values. Maybe change the name to `CheckDidLinkedDomain` and for the values:
- `NEVER`
- `CHECK_IF-AVAILABLE`
- `ALWAYS_CHECK`

sksadjad: changed to:
```
NEVER
IF_PRESENT
ALWAYS
```

sksadjad:
this pr also fixes https://github.com/Sphereon-Opensource/did-auth-siop/issues/21

src/main/functions/LinkedDomainValidations.ts
```diff
@@ -0,0 +1,55 @@
...
+    const messageCondition: boolean = validMessages.includes(err.message);
+    if (checkLinkedDomain === CheckLinkedDomain.ALWAYS || (checkLinkedDomain === CheckLinkedDomain.IF_PRESENT && !messageCondition)) {
+      throw new Error(err.message);
+    }
+  }
+}
+
+async function checkWellknownDid(didDocument: DIDDocument): Promise<IDomainLinkageValidation> {
+  const verifyCallback = async (): Promise<IVerifyCredentialResult> => {
+    return { verified: true };
```
nklomp, line 48: This needs to hook into something which actually verifies the signature. We need to think about whether we want to adopt the same approach as for the well-known DID client itself. Meaning that we leave it up to the user of the lib to choose which VC validation library to use. On the other hand, I think it makes sense to have this library be usable as is. Meaning it already makes some of these choices. Meaning we will have use did-jwt and vc-js from @digitalcredentials to check them depending on the VC type (JWT vs json-ld).

The test code and example in the domain-linkage lib has examples.

test/AuthenticationResponse.response.spec.ts
```diff
@@ -207,10 +212,14 @@ describe('create JWT from Request JWT should', () => {
       responseMode: ResponseMode.POST,
     };
 
+    // nock()
```
nklomp, line 215: Remove

nklomp (commented):
Basically one remark. And that is that it doesn't verify the actual Linked Domain VC.

Since we need to create the code to do actual JWT and JSON-LD VP/VC checking for the AuthResponse payload as well, it might make sense that we create a ticket to do the latter (should already be in Jira), and extend that to also hook it in for the actual signature validation of the Domain Linkage

nklomp (approved):

test/AuthenticationResponse.response.spec.ts
sksadjad: removed

src/main/functions/LinkedDomainValidations.ts
sksadjad: created the ticket

PR #23, opened 2022-09-04
Feature/jose upgrade

nklomp:
- Jose upgrade. Also do thumbprint by jose instead of our own implementation
- Move to new ssi-types package (part of ssi-sdk), to have a common package for VCs/DIDs etc, without any external deps


src/main/PresentationExchange.ts
```diff
@@ -1,19 +1,15 @@
...
   PEX,
   PresentationSignCallBackParams,
   PresentationSignOptions,
-  ProofPurpose,
-  ProofType,
   SelectResults,
   Status,
 } from '@sphereon/pex';
 import { PresentationDefinitionV1, PresentationDefinitionV2, PresentationSubmission } from '@sphereon/pex-models';
+import { IPresentation, IProofPurpose, IProofType, IVerifiableCredential, IVerifiablePresentation } from '@sphereon/ssi-types';
```
sksadjad, line 12: I'm not sure about this, we're adding a dependency to `SSI-SDK` here, but this project is also used by SSI-SDK. And also types can be something separate. maybe we can change the name of `pex-models` to `ssi-models` to include additional things.

src/main/functions/Keys.ts
```diff
@@ -83,16 +67,16 @@ function getThumbprintFromJwkDIDKeyImpl(jwk: JWK): string {
   return `z${bs58.encode(buffer)}`;
 }
 
-export function getThumbprintFromJwk(jwk: JWK, did: string): string {
+export async function getThumbprintFromJwk(jwk: JWK, did: string): Promise<string> {
   if (isEd25519DidKeyMethod(did)) {
     return getThumbprintFromJwkDIDKeyImpl(jwk);
   } else {
-    return getThumbprintFromJwkImpl(jwk);
+    return await calculateJwkThumbprint(jwk, 'sha256');
```
sksadjad, line 74: isn't it better to get it from the request or registration object?

sksadjad (commented):
added small comments but the main thing is, I'm not sure about `@sphereon/ss-types`

src/main/PresentationExchange.ts
nklomp: There is no dep on ssi-sdk itself. Ssi-types has no (transitive) deps. I quite deliberately made that choice to do it like that

src/main/functions/Keys.ts
nklomp: Not sure what you mean. Spec mandates a sha255 thumbprint and this replacement does exactly the same as the previous method it is replacing

src/main/PresentationExchange.ts
nklomp: Also we are not going to use pex-models for that. As that is auto generated code

nklomp: Actually the whole point of having ssi-types and clearing ssi-core was to remove circular deps we had (core depending on siop, which depends on linked domain, which depends on core).

Both core and types now are clean in terms of deps. Meaning the circular deps are goneml. And I want ssi types to be in the mono repo of our sdk

sksadjad (approved):
looks good to me

PR #24, opened 2022-09-08
fix: fixed the issue with populating checkLinkedDomain value

sksadjad:
- fixed the issue with populating checkLinkedDomain value for verifying jwt
- some changes in the validateLinkedDomainWithDid function


src/main/types/SIOP.types.ts
```diff
@@ -452,30 +452,30 @@ export enum VerificationMode {
...
+
+export interface ExternalVerification extends Verification {
   verifyUri: string; // url to call to verify the id_token signature
   authZToken?: string; // Optional: bearer token to use to the call
   resolveOpts: ResolveOpts;
 }
 
 export interface VerifyAuthenticationRequestOpts {
-  verification: InternalVerification | ExternalVerification; // To use internal verification or external hosted verification
+  verification: Verification; // To use internal verification or external hosted verification
```
nklomp, line 470: Don't change this to the super interface. Right now this interface is exactly the same as the InternalVerification. But if we add one unique prop to InternalVerification you would still be able to pass in objects with to few params, as long as they have all props from Verification. It really needs to be either Internal or External Verification

src/main/types/SIOP.types.ts
```diff
@@ -452,30 +452,30 @@ export enum VerificationMode {
...
   checkLinkedDomain?: CheckLinkedDomain;
   // didDocument?: DIDDocument; // If not provided the DID document will be resolved from the request
   nonce?: string; // If provided the nonce in the request needs to match
   // redirectUri?: string;
 }
 
 export interface VerifyAuthenticationResponseOpts {
-  verification: InternalVerification | ExternalVerification;
+  verification: Verification;
```
nklomp, line 478: Don't change this to the super interface. Right now this interface is exactly the same as the InternalVerification. But if we add one unique prop to InternalVerification you would still be able to pass in objects with to few params, as long as they have all props from Verification. It really needs to be either Internal or External Verification

src/main/types/SIOP.types.ts
```diff
@@ -637,9 +637,9 @@ export const isRequestPayload = (object: AuthenticationRequestPayload | Authenti
 export const isResponsePayload = (object: AuthenticationRequestPayload | AuthenticationResponsePayload): object is AuthenticationResponsePayload =>
   'iss' in object && 'aud' in object;
 
-export const isInternalVerification = (object: InternalVerification | ExternalVerification): object is InternalVerification =>
+export const isInternalVerification = (object: Verification): object is InternalVerification =>
```
nklomp, line 640: Don't change this to the super interface. Right now this interface is exactly the same as the InternalVerification. But if we add one unique prop to InternalVerification you would still be able to pass in objects with to few params, as long as they have all props from Verification. It really needs to be either Internal or External Verification

src/main/types/SIOP.types.ts
```diff
@@ -637,9 +637,9 @@ export const isRequestPayload = (object: AuthenticationRequestPayload | Authenti
...
   object.mode === VerificationMode.INTERNAL; /* && !isExternalVerification(object)*/
-export const isExternalVerification = (object: InternalVerification | ExternalVerification): object is ExternalVerification =>
+export const isExternalVerification = (object: Verification): object is ExternalVerification =>
```
nklomp, line 642: Don't change this to the super interface. Right now this interface is exactly the same as the InternalVerification. But if we add one unique prop to InternalVerification you would still be able to pass in objects with to few params, as long as they have all props from Verification. It really needs to be either Internal or External Verification

nklomp (changes requested):
See remark about not using the super interface as argument

src/main/types/SIOP.types.ts
sksadjad: reverted to `InternalVerification | ExternalVerification`

src/main/types/SIOP.types.ts
sksadjad: reverted to `InternalVerification | ExternalVerification`

src/main/types/SIOP.types.ts
sksadjad: reverted to `InternalVerification | ExternalVerification`

src/main/types/SIOP.types.ts
sksadjad: reverted to `InternalVerification | ExternalVerification`

nklomp (approved):

PR #25, opened 2022-09-12
Feature/vdx 137/handle subject syntax types

sksadjad:
updates:
- removed checkedLinkedDomain property from `VerifyAuthenticationResponseOpts` and `VerifyAuthenticationRequestOpts` because we are using it from `Internal/External Verification` object
- [if we don't pass a checkLinkedDomain option to our OP/RP it will use CheckLinkedDomain.IF_PRESENT
- fixed the logic behind finding intersection of OP & RP supported `subjectSyntaxTypes`
- added another enum for `SubjectSyntaxTypesSupportedValues` that has values of `did` and `urn:ietf:params:oauth:jwk-thumbprint`
- fixed the problem with handling the `subject_syntax_types` when OP/RP are initialized using `withDidMethod`
- fixed the problem with `nonce` value in populating `AuthenticationRequestPayload`
- fixed the tests with the correct logic of finding intersection of of OP & RP supported `subjectSyntaxTypes`

### DIFF
--- a/src/main/AuthenticationRequest.ts
+++ b/src/main/AuthenticationRequest.ts
@@ -275,7 +275,6 @@ function createClaimsPayload(opts: ClaimOpts, nonce: string): ClaimPayload {
           throw new Error(SIOPErrors.REQUEST_CLAIMS_PRESENTATION_DEFINITION_NOT_VALID);
         } else {
           vp_token = {
-            //TODO: nonce should be initialized correctly
             nonce: nonce,
             presentation_definition: def.definition,
             response_type: PresentationLocation.VP_TOKEN,
--- a/src/main/AuthenticationRequest.ts
+++ b/src/main/AuthenticationRequest.ts
@@ -134,6 +134,8 @@ export default class AuthenticationRequest {
     }
     if (opts.verification.checkLinkedDomain && opts.verification.checkLinkedDomain != CheckLinkedDomain.NEVER) {
       await validateLinkedDomainWithDid(verPayload.iss, opts.verification.checkLinkedDomain);
+    } else if (!opts.verification.checkLinkedDomain) {
+      await validateLinkedDomainWithDid(verPayload.iss, CheckLinkedDomain.IF_PRESENT);
     }
     const presentationDefinitions = await PresentationExchange.findValidPresentationDefinitions(payload);
     return {
--- a/src/main/AuthenticationRequest.ts
+++ b/src/main/AuthenticationRequest.ts
@@ -248,7 +248,7 @@ function assertValidRequestOpts(opts: AuthenticationRequestOpts) {
   assertValidRequestRegistrationOpts(opts.registration);
 }
 
-function createClaimsPayload(opts: ClaimOpts): ClaimPayload {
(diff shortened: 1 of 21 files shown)


src/main/AuthenticationRequest.ts
```diff
@@ -276,7 +276,7 @@ function createClaimsPayload(opts: ClaimOpts): ClaimPayload {
         } else {
           vp_token = {
             //TODO: nonce should be initialized correctly
-            nonce: 'NONCE_STRING',
+            nonce: nonce,
```
nklomp, line 279: could be just nonce. Als remove the TODO

src/main/AuthenticationResponse.ts
```diff
@@ -268,6 +267,12 @@ async function createSIOPResponsePayload(
     vp_token,
     verifiable_presentations,
   };
+  if (supportedDidMethods.length && resOpts.did) {
```
nklomp, line 270: Hmzz, this doesn't sound right.
So if we have a DID we are using a jwk/thumbprint. Should be the other way around

src/main/OP.ts
```diff
@@ -32,7 +32,7 @@ export class OP {
 
   public constructor(opts: { builder?: OPBuilder; responseOpts?: AuthenticationResponseOpts; verifyOpts?: VerifyAuthenticationRequestOpts }) {
     this._authResponseOpts = { ...createResponseOptsFromBuilderOrExistingOpts(opts) };
-    this._verifyAuthRequestOpts = { ...createVerifyRequestOptsFromBuilderOrExistingOpts(opts) };
+    this._verifyAuthRequestOpts = { ...createVerifyRequestOptsFromBuilderOrExistingOpts(opts) } as Partial<VerifyAuthenticationRequestOpts>;
```
nklomp, line 35: What is the reason we need a partial all of a sudden. Sounds like something is off

src/main/types/SSI.types.ts
```diff
@@ -3,6 +3,7 @@ import { JWK } from 'jose';
 
 export interface ResolveOpts {
   resolver?: Resolvable;
+  resolvers?: Map<string, Resolvable>;
```
nklomp, line 6: TBH I am not sure about this approach. Probably we should place a todo here linking to the ticket recently created

nklomp (changes requested):
Some small remarks

src/main/AuthenticationRequest.ts
sksadjad: todo removed

src/main/AuthenticationResponse.ts
sksadjad: yes, this should be other way around

src/main/OP.ts
sksadjad: yes, changed some upper method calls expected a partial when it fact they have received a full object. no need for this, removed

src/main/types/SSI.types.ts
sksadjad: added todo for the other ticket

nklomp (approved):
LGTM

PR #26, opened 2022-09-14
Feature/vdx 135/verify credential

maikel-maas:


package.json
```diff
@@ -27,10 +27,14 @@
     "node": ">=14"
   },
   "dependencies": {
+    "@digitalcredentials/ed25519-signature-2020": "^3.0.2",
+    "@digitalcredentials/vc": "^4.1.1",
     "@sphereon/did-uni-client": "^0.4.0",
-    "@sphereon/ssi-types": "^0.8.1-next.3",
+    "@sphereon/jsonld-vc-verifier": "./sphereon-jsonld-vc-verifier-v1.0.0.tgz",
```
nklomp, line 33: I don't want this whole christmas tree as deps. Should be handled using callbacks like statuslist

nklomp (changes requested):
I don't want this whole christmas tree as deps. Should be handled using callbacks like statuslist and thus add any deps as dev depencies and use tests to show how it would work with callbacks

maikel-maas:
Closing this PR based on the comments above

PR #28, opened 2022-09-20
WIP: fix resolver configuration

sksadjad:


src/main/RP.ts
```diff
@@ -28,6 +28,7 @@ export class RP {
     const claims = opts.builder?.claims;
     this._authRequestOpts = { claims, ...createRequestOptsFromBuilderOrExistingOpts(opts) };
     this._verifyAuthResponseOpts = { claims, ...createVerifyResponseOptsFromBuilderOrExistingOpts(opts) };
+    console.log('RP:31: ', this._verifyAuthResponseOpts);
```
nklomp, line 31: Remove console.log

src/main/OP.ts
```diff
@@ -246,13 +246,9 @@ function createVerifyRequestOptsFromBuilderOrExistingOpts(opts: {
...
-            resolver: opts.builder.resolver,
-            resolvers:
-              //TODO: discuss this with Niels
-              //https://sphereon.atlassian.net/browse/VDX-139
-              opts.builder.resolvers && opts.builder.resolvers.size
-                ? opts.builder.resolvers
-                : getResolver({ subjectSyntaxTypesSupported: subjectSyntaxTypesSupported }),
+            resolver: opts.builder.customResolver
+              ? opts.builder.customResolver
+              : getResolverUnion(opts.builder.responseRegistration.subjectSyntaxTypesSupported, opts.builder.resolvers),
```
nklomp, line 251: I think we need to make some changes, as the handling is onconsistent. What needs to happen is that we change this method for instance:

````typescript
 addDidMethod(didMethod: string, opts?: { resolveUrl?: string; baseUrl?: string }): OPBuilder {
    if (didMethod.startsWith('did:')) {
      this.addResolver(getMethodFromDid(didMethod), new Resolver(getUniResolver(getMethodFromDid(didMethod), { ...opts })));
      this.didMethods.push(getMethodFromDid(didMethod));
    } else {
      this.addResolver(didMethod, new Resolver(getUniResolver(didMethod, { ...opts })));
      this.didMethods.push(didMethod);
    }
    return this;
  }
````
As you can see, if you now call addDidMethod it will always create a new uniresolver for it. Which is similar to what you are doing in the getResolverUnion. The easy fix would be to just allow the map to have a null/undefined resolvable as value in the resolvers property, so that you indeed can handle it in your code


Also addResolver. directly manipulates the subjectSyntaxTypes supported it seems, whilst addDidMethod first populates the 'didMethods' property, which then gets merged into the subjectSyntaxTypes when the buildder builds the OP. We need to do one, or the other, but not mix them

**Review by nklomp — CHANGES_REQUESTED** [review]
See remarks.

Probably it makes sense to introduce something like a resolver/did registry, which can be used from both the RP and OP sides. There are similar methods in both builders, where only the argument names differ slightly. It would probably be better to create a registry instance in the OP/RP Builder and then delegate adding DID methods and DID methods with specific resolvers, as well as the custom resolver to that registry instance. Then we can just copy that instance to the OP resp RP when the build() method is called. That ensures we have a common shared code base to handle the DIDs supported and their resolvers and it ensures we will not introduce bugs by forgetting to keep the RP and OP sides in sync when we make changes

File: src/main/OP.ts (hunk 2 above)
Comment (sksadjad) [reply src/main/OP.ts:251]: removed the `didMethods` from the builder files. now we just populate the `Map<string, Resolvable>` and then later on, in the actual build of the OP/RP, we add the did values from the map to `subjectSyntaxTypes`.
But I also have an unrelated questions about this. what is the correct form of a did method? let me rephrase that, what do we want as the end result of our process?
1. `example`
2. `did:example`
3. `did:example:`

right now I have changed the end result to be `did:example` I'm asking because we have a method in `DidJWT` called `toSIOPRegistrationDidMethod` which produces `did:example:`

File: src/main/RP.ts (hunk 1 above)
Comment (sksadjad) [reply src/main/RP.ts]: removed

File: src/main/OP.ts (hunk 2 above)
Comment (nklomp) [reply src/main/OP.ts:248]: As mentioned. The SIOPv2 spec incorrectly has did:example as a DID method (it was worse when they had did:example:  as DID method). The actual DID method is only the 'example' portion of it (see DID core spec). 
So as long as we make sure that we accept did:example, example or even did:example:1234567 then correct that to a correct DID method (example) internally. Then whenever we start populating a SIOP payload, we have to adjust to did:example. The toSIOPRegistrationDidMethod is meant for this (it adds the last colon, because that was in the spec back then)

File: src/main/functions/DIDResolution.ts [hunk 3]
```diff
@@ -31,6 +31,42 @@ export function getResolver(opts: ResolveOpts): Resolvable {
   }
 }
 
+export function getResolverUnion(subjectSyntaxTypesSupported: string[] | string, resolverMap: Map<string, Resolvable>): Resolvable {
```
nklomp, line 34: I rather would have seen this as part of a class/object that also contains the map on the RP/OP builder side

src/main/functions/DidJWT.ts
```diff
@@ -273,3 +273,12 @@ export function toSIOPRegistrationDidMethod(didOrMethod: string) {
   const split = prefix.split(':');
   return `${split[0]}:${split[1]}:`;
 }
+
+export function toQualifiedDidMethod(didOrMethod: string) {
```
nklomp, line 277: What does qualified mean? Because the end result is not a DID method according to DID-core. That is why there is the toSiopRegistrationDidMethod. That method needs this body basically (so removing the last : )

nklomp (changes requested):
Some remarks on the PR, and a seperate comment on your question about DID methods.

Main point is that I really would like to see the DID map and the separate function move to an Object

src/main/functions/DIDResolution.ts
```diff
@@ -31,6 +31,42 @@ export function getResolver(opts: ResolveOpts): Resolvable {
...
+    [p: string]: (did: string, _parsed: ParsedDID, _didResolver: Resolver, _options: DIDResolutionOptions) => Promise<DIDResolutionResult>;
+  }[] = [];
+  const subjectTypes: string[] = [];
+  if (subjectSyntaxTypesSupported) {
+    typeof subjectSyntaxTypesSupported === 'string'
+      ? subjectTypes.push(subjectSyntaxTypesSupported)
+      : subjectTypes.push(...subjectSyntaxTypesSupported);
+  }
+  if (subjectTypes.indexOf(SubjectSyntaxTypesSupportedValues.DID.valueOf()) !== -1) {
+    return new UniResolver();
```
nklomp, line 45: This should be the customResolver if it is set btw, otherwise the UnionResolver.

So the first lines in getResolver which always returns the customResolver if set, is wrong. 

We need to populate the map values first. In case a resolver is set for the DID method, we will use that. In case a customResolver is set and the value is null we will use the customResolver for that method. In case no customResolver is set and the value is null, we will use the UniResolver

nklomp (changes requested):
Also a comment on the actual implementation. That is not really in line with the ticket.

Use a customResolver per DID method if the value in the map is null and the custom resolver is set. Ohterwise use the uni resolver in case the value in the map is null

Do not default to the customResolver for everything in case it is set.
Of someone wants that, they could simply set the customResolver and then only call addDidMethod

src/main/functions/DIDResolution.ts
sksadjad: changed to return the `customResolver` in the first line, the rest is changed according to our slack discussion

src/main/functions/DidJWT.ts
sksadjad: removed this method and changed the impl of `toSiopRegistrationDidMethod` to return `did:example`

src/main/functions/DIDResolution.ts
sksadjad: changed the impl of `getResolverUnion`. now we have two separate method: `mergeAllDidMethods` which returns the content that should end up in `subject_syntax_types_supported`. But having this in OP/RP will force us to duplicate codes for this

src/main/OPBuilder.ts
```diff
@@ -22,19 +22,15 @@ export default class OPBuilder {
...
-    if (didMethod.startsWith('did:')) {
-      this.addResolver(getMethodFromDid(didMethod), new Resolver(getUniResolver(getMethodFromDid(didMethod), { ...opts })));
-      this.didMethods.push(getMethodFromDid(didMethod));
-    } else {
-      this.addResolver(didMethod, new Resolver(getUniResolver(didMethod, { ...opts })));
-      this.didMethods.push(didMethod);
-    }
+    const qualifiedDidMethod = didMethod.startsWith('did:') ? getMethodFromDid(didMethod) : didMethod;
+    opts
+      ? this.addResolver(getMethodFromDid(didMethod), new Resolver(getUniResolver(qualifiedDidMethod, { ...opts })))
```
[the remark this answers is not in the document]
sksadjad: Since I don't handle this anymore, I'll add what you asked for in the `getResolverUnion` which is handling the `resolver` property of both OP/RP

src/main/OP.ts
sksadjad: fixed this according to our talk

src/main/functions/DIDResolution.ts
```diff
@@ -31,6 +31,63 @@ export function getResolver(opts: ResolveOpts): Resolvable {
...
+      : subjectTypes.push(...subjectSyntaxTypesSupported);
+  }
+  if (subjectTypes.indexOf(SubjectSyntaxTypesSupportedValues.DID.valueOf()) !== -1) {
+    return customResolver ? customResolver : new UniResolver();
+  }
+  const specificDidMethods = subjectTypes.filter((sst) => !!sst && sst.startsWith('did:'));
+  specificDidMethods.forEach((dm) => {
+    let uniResolver;
+    if (!resolverMap.has(dm)) {
+      uniResolver = getUniResolver(getMethodFromDid(dm));
```
nklomp, line 70: I still would expect customResolver here if defined. Now the user is saying. This is my custom resolver. The implementation is doing. Oh see, we have a did method without a resolver, let's use the uni resolver. That is almost certainly not what the user intended to happen.

src/main/functions/DIDResolution.ts
```diff
@@ -31,6 +31,63 @@ export function getResolver(opts: ResolveOpts): Resolvable {
...
+    } else {
+      uniResolver = getUniResolver(getMethodFromDid(dm), resolverMap.get(dm));
```
nklomp, line 72: Why are we always using a uni resolver?

nklomp (changes requested):
See remarks

src/main/OP.ts
```diff
@@ -225,34 +223,24 @@ function createVerifyRequestOptsFromBuilderOrExistingOpts(opts: {
...
-      subjectSyntaxTypesSupported.push(opts.builder.responseRegistration.subjectSyntaxTypesSupported);
-    }
-  } else if (opts?.verifyOpts?.verification?.resolveOpts?.subjectSyntaxTypesSupported) {
-    if (Array.isArray(opts.verifyOpts.verification.resolveOpts.subjectSyntaxTypesSupported)) {
-      subjectSyntaxTypesSupported.push(...opts.verifyOpts.verification.resolveOpts.subjectSyntaxTypesSupported);
-    } else {
-      subjectSyntaxTypesSupported.push(opts.verifyOpts.verification.resolveOpts.subjectSyntaxTypesSupported);
-    }
+  if (opts?.builder?.resolvers.size && opts.builder?.responseRegistration) {
+    opts.builder.responseRegistration.subjectSyntaxTypesSupported = mergeAllDidMethods(
```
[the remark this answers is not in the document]
sksadjad: yeah I struggled with it as well. I thought since our user has given us a registeration object, and also provided us with some did methods, I have to add that to the `subject_syntax_types_supported` property of the registeration object. because in the negotiation process, we want to express all the supported methods

src/main/RP.ts
```diff
@@ -118,15 +119,16 @@ function createRequestOptsFromBuilderOrExistingOpts(opts: { builder?: RPBuilder;
...
-    if (!Array.isArray(opts.builder.requestRegistration.subjectSyntaxTypesSupported)) {
-      opts.builder.requestRegistration.subjectSyntaxTypesSupported = [opts.builder.requestRegistration.subjectSyntaxTypesSupported];
-    }
-    const unionSubjectSyntaxTypes = new Set();
-    opts.builder.requestRegistration.subjectSyntaxTypesSupported.forEach((sst) => unionSubjectSyntaxTypes.add(sst));
-    opts.builder.didMethods.forEach((didMethod) => unionSubjectSyntaxTypes.add(toSIOPRegistrationDidMethod(didMethod)));
-    opts.builder.requestRegistration.subjectSyntaxTypesSupported = Array.from(unionSubjectSyntaxTypes) as string[];
+function createVerifyResponseOptsFromBuilderOrExistingOpts(opts: { builder?: RPBuilder; verifyOpts?: VerifyAuthenticationResponseOpts }) {
+  if (opts?.builder?.resolvers.size && opts.builder?.requestRegistration) {
+    opts.builder.requestRegistration.subjectSyntaxTypesSupported = mergeAllDidMethods(
```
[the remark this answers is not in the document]
sksadjad: same as above

src/main/functions/DIDResolution.ts
sksadjad: yeah this is exactly what is happenning. look at line 50

src/main/functions/DIDResolution.ts
sksadjad: we're using UniResolver with the config provided from the user

nklomp: But a difResolver is not a uni resolver. The uni resolver delegates to something hosted, whilst a dif resolver could be resolving directly for instance.

sksadjad: changed it to:
```
 if (!resolverMap.has(dm) || resolverMap.get(dm) === null) {
      methodResolver = getUniResolver(getMethodFromDid(dm));
    } else {
      methodResolver = resolverMap.get(dm);
    }
```

src/main/functions/DIDResolution.ts
```diff
@@ -31,6 +31,63 @@ export function getResolver(opts: ResolveOpts): Resolvable {
...
+    : new Resolver(...uniResolvers);
+}
+
+export function mergeAllDidMethods(subjectSyntaxTypesSupported: string | string[], resolvers: Map<string, Resolvable>): string[] {
+  if (!Array.isArray(subjectSyntaxTypesSupported)) {
+    subjectSyntaxTypesSupported = [subjectSyntaxTypesSupported];
+  }
+  const unionSubjectSyntaxTypes = new Set();
+  subjectSyntaxTypesSupported.forEach((sst) => unionSubjectSyntaxTypes.add(sst));
+  resolvers.forEach((_value, didMethod) => unionSubjectSyntaxTypes.add(toSIOPRegistrationDidMethod(didMethod)));
```
maikel-maas, line 87: If you are not using the _value, replace it with _

sksadjad: changed to `_`

src/main/OP.ts
```diff
@@ -154,14 +155,11 @@ async function parseAndResolveUri(encodedUri: string) {
...
-  if (opts?.builder?.didMethods.length && opts.builder?.responseRegistration?.subjectSyntaxTypesSupported) {
-    if (!Array.isArray(opts.builder.responseRegistration.subjectSyntaxTypesSupported)) {
-      opts.builder.responseRegistration.subjectSyntaxTypesSupported = [opts.builder.responseRegistration.subjectSyntaxTypesSupported];
-    }
-    const unionSubjectSyntaxTypes = new Set();
-    opts.builder.responseRegistration.subjectSyntaxTypesSupported.forEach((sst) => unionSubjectSyntaxTypes.add(sst));
-    opts.builder.didMethods.forEach((didMethod) => unionSubjectSyntaxTypes.add(toSIOPRegistrationDidMethod(didMethod)));
-    opts.builder.responseRegistration.subjectSyntaxTypesSupported = Array.from(unionSubjectSyntaxTypes) as string[];
+  if (opts?.builder?.resolvers.size && opts.builder?.responseRegistration?.subjectSyntaxTypesSupported) {
+    opts.builder.responseRegistration.subjectSyntaxTypesSupported = mergeAllDidMethods(
```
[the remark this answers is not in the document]
sksadjad: this is a side thing to the tocket. here I'm indeed merging did methods. populating the resolver is happening in DIDResolution

sksadjad:
> In general it looks good to me, can we improve the code coverage?

in th scope of this ticket, I don't think so. this is just for improving and unifying the way we're handling the resolver object

src/main/RP.ts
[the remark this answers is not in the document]
nklomp: Let's leave it, but make a comment that we might want to update in the future

src/main/functions/DIDResolution.ts
nklomp: Yeah, that is not entirely what I had in mind. But let's leave it for now. I wanted the customResolver to be a fallback, but still allow a DID method to use another resolver. Current implementation just uses customResolver for everything if it is set.

I will create another ticket during the refactor tickets, for this

nklomp (approved):

nklomp:
See remark to place 2 remarks in the code

PR #30, opened 2022-09-22
VDX-104

BtencateSphereon:
Uses local dep version, needs to be updated before merging

will fix the conflicts when dependencies can be updated


maikel-maas (commented):
In general it looks good, however it needs to be rebased and the dependencies need to be updated.

sksadjad (changes requested):
overall looks good. but two points:
1. build is failing because of using `W3CVerifiableCredential` which is not exported in `ssi-types`
2. run a `yarn fix`

sksadjad (approved):
looks good to me

PR #32, opened 2022-09-27
Feature/vdx 149/profiles enum

hrehman-sphereon:


sksadjad (changes requested):
overall looks good. two small notes, `profile` seems a bit off, because we're talking about different versions of the SIOP protocol here. and two remove two of the three methods you've added: `withProfileStr`, `addProfile`, `withProfiles`. and make the remaining one accept both list and singular value

hrehman-sphereon:
```

public withProfiles(profiles: string | Profile | Array<Profile>): OPBuilder {
    if(typeof profiles === "string") {
      this.withProfileStr(profiles);
    } else if (profiles instanceof Profile) {
      this.addProfile(profiles as unknown as Profile);
    } else if (Array.isArray(profiles)) {
      this.withProfilesList(profiles);
    }

    return this;
  }
```
  
  `profiles instanceof Profile` .... actually have to be implemented like so ... `Array.isArray(profiles)`
  
  
  https://stackoverflow.com/questions/53672528/using-instanceof-with-typescript-union-types-customobject-customobject
  
  
This does not seem an elegant solution.

hrehman-sphereon:
**This document is not a specification, but a profile. It outlines existing specifications required for implementations to interoperate among each other.**

Quoted from : https://identity.foundation/jwt-vc-presentation-profile/

I think the word "profile" does work... do you have a better word in mind ?

src/main/types/SupportedVersion.ts
```diff
@@ -0,0 +1,5 @@
+export enum SupportedVersion {
```
sksadjad, line 1: add this to our SIOP types

test/OPBuilder.spec.ts
```diff
@@ -0,0 +1,48 @@
+import { IVerifyCallbackArgs, IVerifyCredentialResult } from '@sphereon/wellknown-dids-client';
```
sksadjad, line 1: we have a couple of builder tests already in the `OP.request.spec.ts` please move these to that test suite

test/RPBuilder.spec.ts
```diff
@@ -0,0 +1,50 @@
+import { IVerifyCallbackArgs, IVerifyCredentialResult } from '@sphereon/wellknown-dids-client';
```
sksadjad, line 1: we have a couple of builder tests already in the `RP.request.spec.ts` please move these to that test suite

sksadjad (changes requested):
getting there, some small changes requested

src/main/types/SupportedVersion.ts
hrehman-sphereon: fixed

test/OPBuilder.spec.ts
hrehman-sphereon: fixed

test/RPBuilder.spec.ts
hrehman-sphereon: fixed

sksadjad (approved):
looks good to me

PR #33, opened 2022-09-27
Feature/vdx 138/verify vp

maikel-maas:
Extracted JWT verification to callback function;


maikel-maas:
I will open another PR.

PR #34, opened 2022-09-28
Feature/vdx 148/add verifier information for wallet clients

hrehman-sphereon:


sksadjad (changes requested):
looks good to me, just something I forgot to mention yesterday, please add some assertion for the new added values in the RP and OP test suits

hrehman-sphereon:
> looks good to me, just something I forgot to mention yesterday, please add some assertion for the new added values in the RP and OP test suits

Some assertions added, for example, in the jest case 'return a reference url'

nklomp:
As mentioned in a comment I would expect a builder that has a seperate optional argument for a language tag and a required argument. The builder than builds the complete tags

hrehman-sphereon:
> As mentioned in a comment I would expect a builder that has a seperate optional argument for a language tag and a required argument. The builder than builds the complete tags

@nklomp 

test/IT.spec.ts
> 'RP and OP interaction should'   'succeed when calling each other in the full flow'
line 149.

From this same example we do allow user to set the variable. We do not have a builder method for eavery single field of the registration object. Would you like to have builder method for these 3 fields or all fields of registration ?

hrehman-sphereon:
> The builder than builds the complete tags

What does it mean ? we do not have field for every single language possible (cross product) and for every single field. So ... do we need to get rid of `[x: string]: any` and add fixed fields ? I hope this will be against the RFC as well as I don't think that's what you mean...

so, can you please explain ?

src/main/LanguageTagUtils.ts
```diff
@@ -0,0 +1,128 @@
...
+          const value = entry[1];
+          if (this.isStringNullOrEmpty(key) || this.isStringNullOrEmpty(value)) {
+            throw new Error(SIOPErrors.BAD_PARAMS + '. languageTagEnabledFieldsName must be non-null or non-empty');
+          }
+        }
+      }
+    }
+  }
+
+  private static isStringNullOrEmpty(key: string) {
```
sksadjad, line 119: move this into `ObjectUtils`

sksadjad (changes requested):
added two small notes, otherwise looks good

src/main/LanguageTagUtils.ts
```diff
@@ -0,0 +1,128 @@
+import Tags from 'language-tags';
+
+import { SIOPErrors } from './types';
+
+export default class LanguageTagUtils {
```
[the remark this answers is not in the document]
hrehman-sphereon: fixed

src/main/LanguageTagUtils.ts
hrehman-sphereon: fixed

sksadjad (approved):
approved

PR #35, opened 2022-09-28
VDX-138 Check the signature of VC and VPs received in RP side

maikel-maas:
- Added VP verification callback;


sksadjad (changes requested):
We should be able to configure OP/RP, with 3 options: NEVER, IF_PRESENT, ALWAYS (if not configured-> IF_PRESENT), and based on this value, you'll decide if you want to assert or not

nklomp:
No. VP checks should always be performed

sksadjad (approved):
looks good to me

sksadjad (changes requested):
build is failing

PR #37, opened 2022-09-29
VDX-144 Adjust the SIOPv2 lib documenation with latest changes

maikel-maas:
- Updated README file;


nklomp:
I hate documents with all kinds of pointers to other documents in Github

nklomp:
I rather have a future documentation website like for VDX, that does proper chapter, headers and links, versus manually linking Markdown files on Github. Not sure how this requirement became part of the ticket

hrehman-sphereon (approved):

hrehman-sphereon (changes requested):
There are some code example that I will fix.

sksadjad (commented):
TBH I don't think we can merge this PR yet. the impl is changing as a result of covering different versions like ID1 and ID11. we also have some additional tickets that will result in further changes in the way we're building OP/RP.

hrehman-sphereon (approved):
Approving as we just need to merge whatever is currently present.

Agreed with @sksadjad

PR #38, opened 2022-10-03
Feature/vdx 156/add type defs

hrehman-sphereon:


src/main/PresentationExchange.ts
```diff
@@ -130,19 +130,19 @@ export class PresentationExchange {
     const allDefinitions: PresentationDefinitionWithLocation[] = [];
 
     async function extractPDFromVPToken() {
-      const vpTokens = extractDataFromPath(obj, '$..vp_token.presentation_definition');
-      const vpTokenRef = extractDataFromPath(obj, '$..vp_token.presentation_definition_uri');
-      if (vpTokens && vpTokens.length && vpTokenRef && vpTokenRef.length) {
+      const vpTokens: PresentationDefinitionV1[] | PresentationDefinitionV2[] = extractDataFromPath(obj, '$..vp_token.presentation_definition');
```
sksadjad, line 133: since we're assuming the type here, can you add a test case for this?

src/main/PresentationExchange.ts
```diff
@@ -151,16 +151,16 @@ export class PresentationExchange {
...
-      }
-      allDefinitions.push({ definition: definition.value, location: PresentationLocation.ID_TOKEN });
+    function addSingleIdTokenPDToPDs(definition: IPresentationDefinition): void {
+      PresentationExchange.assertValidPresentationDefinition(definition);
+      allDefinitions.push({ definition: definition, location: PresentationLocation.ID_TOKEN });
     }
 
     async function extractPDFromOtherTokens() {
-      const definitions = extractDataFromPath(obj, '$..verifiable_presentations.presentation_definition');
+      const definitions: PresentationDefinitionV1[] | PresentationDefinitionV2[] = extractDataFromPath(
```
sksadjad, line 160: same as above, can you add a test case for this?

sksadjad (changes requested):
looks good, but I'm not sure about two types, can you add test case for those?

src/main/PresentationExchange.ts
maikel-maas: Without an explicit check it will not help, at runtime it can be anything.

PR #40, opened 2022-10-06
feature/VDX-157_sign_vp_callback

hrehman-sphereon:


sksadjad:
Added usecases for `PresentationSignCallback` @nklomp can you take a look and see if you agree with the way I'm handling it?

sksadjad:
> In this repo it looks okay. Have to check the actual PresentationSign callback in PEX though

it's ok with the latest pex, but it's a local build

src/main/PresentationExchange.ts
```diff
@@ -1,21 +1,13 @@
...
-  EvaluationResults,
-  IPresentationDefinition,
-  KeyEncoding,
-  PEX,
-  PresentationSignCallBackParams,
-  PresentationSignOptions,
-  SelectResults,
-  Status,
-} from '@sphereon/pex';
+import { EvaluationResults, IPresentationDefinition, KeyEncoding, PEX, PresentationSignOptions, SelectResults, Status } from '@sphereon/pex';
```
nklomp, line 1: Github gives these: 

> Argument of type 'PresentationSignCallback' is not assignable to parameter of type '(callBackParams: PresentationSignCallBackParams) => Promise<IVerifiablePresentation>'.

sksadjad: yeah, as I said it's a local build of pex, I didn't push the local pex library that I'm using

nklomp (approved):
LGTM

PR #41, opened 2022-10-10
changed the logic for id_token to contain PS as _vp_token

sksadjad:
- contains changes in constructing id_token and vp_token (and how we process them using pex)
- also changed the ethr dids that were failing to resolve

### DIFF
--- a/src/main/AuthenticationResponse.ts
+++ b/src/main/AuthenticationResponse.ts
@@ -290,6 +290,7 @@ async function createSIOPResponsePayload(
     exp: Date.now() / 1000 + (resOpts.expiresIn || 600),
     registration,
   };
+  // add support for multiple VPs (VDX-158)
   if (resOpts.vp && resOpts.vp[0].location === PresentationLocation.ID_TOKEN) {
     const id_token = {
       iss: ResponseIss.SELF_ISSUED_V2,
--- a/src/main/AuthenticationResponse.ts
+++ b/src/main/AuthenticationResponse.ts
@@ -1,3 +1,4 @@
+import { ObjectUtils, PresentationSubmission } from '@sphereon/ssi-types';
 import { JWTHeader } from 'did-jwt';
 import { JWK } from 'jose';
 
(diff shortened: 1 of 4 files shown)


test/PresentationExchange.spec.ts
```diff
@@ -370,7 +370,7 @@ describe('presentation exchange manager tests', () => {
       format: VerifiablePresentationTypeFormat.LDP_VP,
     };
     try {
-      await PresentationExchange.validatePayloadsAgainstDefinitions(pd, [vpw]);
+      await PresentationExchange.validatePayloadsAgainstDefinitions(pd, [vpw], undefined);
```
nklomp, line 373: Why do we need the undefined here?

src/main/AuthenticationResponse.ts
```diff
@@ -288,14 +289,57 @@ async function createSIOPResponsePayload(
...
+      iss: ResponseIss.SELF_ISSUED_V2,
+      aud: verifiedJwt.payload.redirect_uri,
+      iat: Date.now() / 1000,
+      exp: Date.now() / 1000 + (resOpts.expiresIn || 600),
+      sub: resOpts.did,
+      //TODO: We don't have this value in the payload
+      auth_time: verifiedJwt.payload.auth_time,
+      nonce: nonce,
+      _vp_token: {
+        // right now pex doesn't support presentation_submission for multiple VPs (as well as multiple VPs)
```
nklomp, line 304: Can we create a ticket that we refine later in Jira and link to it

nklomp (commented):
Some small remarks

src/main/AuthenticationResponse.ts
sksadjad: created 3:
1. to support `presentation_submission` outside of VP in PEX
2. to support multiple VPs in PEX
3. to handle multiple VPs in this project (did-auth-siop)

test/PresentationExchange.spec.ts
sksadjad: made it optional

PR #49, opened 2023-02-08
[VDX-165] update the version

hrehman-sphereon:


nklomp (changes requested):
Not gonna approve. Console.logs all over the place and the more obscure errors are very much on purpose for a lib that does auth

PR #50, opened 2023-02-13
VDX-162

BtencateSphereon:


src/main/op/OP.ts
```diff
@@ -14,13 +16,17 @@ import {
...
+  SIOPResonse,
   UrlEncodingFormat,
   VerifiedAuthorizationRequest,
 } from '../types';
+import { AuthorizationEvents } from '../types/Events';
 
 import { Builder } from './Builder';
 import { createResponseOptsFromBuilderOrExistingOpts, createVerifyRequestOptsFromBuilderOrExistingOpts } from './Opts';
 
+export const opEventEmitter = new EventEmitter();
```
nklomp, line 28: Why not create this in the OP? You then have a new emitter per OP

src/main/op/OP.ts
```diff
@@ -59,8 +65,22 @@ export class OP {
     requestJwtOrUri: string | URI,
     requestOpts?: { nonce?: string; verification?: InternalVerification | ExternalVerification }
   ): Promise<VerifiedAuthorizationRequest> {
-    const authorizationRequest = await AuthorizationRequest.fromUriOrJwt(requestJwtOrUri);
-    return await authorizationRequest.verify(this.newVerifyAuthorizationRequestOpts({ ...requestOpts }));
+    const authorizationRequest = await AuthorizationRequest.fromUriOrJwt(requestJwtOrUri).catch((error: Error) => {
+      opEventEmitter.emit(AuthorizationEvents.ON_AUTH_REQUEST_RECEIVED_FAILED, requestJwtOrUri, error);
```
nklomp, line 69: Could you change the order to be more similar to the one below.

So a then and then a catch, instead of a catch en then having the emit outside of the flow

src/main/rp/RP.ts
```diff
@@ -18,10 +20,13 @@ import {
   SupportedVersion,
   VerifiedAuthenticationResponse,
 } from '../types';
+import { AuthorizationEvents } from '../types/Events';
 
 import Builder from './Builder';
 import { createRequestOptsFromBuilderOrExistingOpts, createVerifyResponseOptsFromBuilderOrExistingOpts, isTargetOrNoTargets } from './Opts';
 
+export const rpEventEmitter = new EventEmitter();
```
nklomp, line 28: Move to RP so we have an emitter per RP (you can have multiple RPs if you want)

src/main/rp/RP.ts
```diff
@@ -98,9 +123,22 @@ export class RP {
     const verifyAuthenticationResponseOpts = this.newVerifyAuthorizationResponseOpts({
       ...opts,
     });
-    const authorizationResponse = await AuthorizationResponse.fromPayload(authorizationResponsePayload);
-
-    return await authorizationResponse.verify(verifyAuthenticationResponseOpts);
+    const authorizationResponse = await AuthorizationResponse.fromPayload(authorizationResponsePayload).catch((error: Error) => {
```
nklomp, line 126: then, catch instead of catch and emit outside of the flow. So similar to the approach done below

src/main/rp/ReplayRegistry.ts
```diff
@@ -0,0 +1,126 @@
...
+  AuthorizationResponseStateStatus,
+} from '../types/ReplayRegistry';
+
+import { rpEventEmitter } from './RP';
+
+export class ReplayRegistry {
+  private authorizationRequests: Map<string, AuthorizationRequestState> = new Map<string, AuthorizationRequestState>();
+  private authorizationResponses: Map<string, AuthorizationResponseState> = new Map<string, AuthorizationResponseState>();
+
+  public constructor() {
```
nklomp, line 18: if you have the emitter as constructor argument, you can have multiple replay registries (one per RP)

src/main/rp/ReplayRegistry.ts
```diff
@@ -0,0 +1,126 @@
...
+    rpEventEmitter.on(AuthorizationEvents.ON_AUTH_REQUEST_CREATED_SUCCESS, this.onAuthorizationRequestCreatedSuccess.bind(this));
+    rpEventEmitter.on(AuthorizationEvents.ON_AUTH_RESPONSE_RECEIVED_SUCCESS, this.onAuthorizationResponseReceivedSuccess.bind(this));
+    rpEventEmitter.on(AuthorizationEvents.ON_AUTH_RESPONSE_VERIFIED_SUCCESS, this.onAuthorizationResponseVerifiedSuccess.bind(this));
+    rpEventEmitter.on(AuthorizationEvents.ON_AUTH_RESPONSE_VERIFIED_FAILED, this.onAuthorizationResponseVerifiedFailed.bind(this));
+  }
+
+  private async onAuthorizationRequestCreatedSuccess(authorizationRequest: AuthorizationRequest): Promise<void> {
+    try {
+      if (!authorizationRequest) {
+        return;
```
nklomp, line 28: Hmzz, I think we might want to throw an error

src/main/rp/ReplayRegistry.ts
```diff
@@ -0,0 +1,126 @@
...
+      }
+
+      const timestamp = Date.now();
+      const payload: RequestObjectPayload | undefined = await authorizationRequest.requestObject.getPayload();
+      if (!payload) {
+        throw new Error('Request does not contain a payload');
+      }
+      this.authorizationRequests.set(payload.nonce, { payload, status: AuthorizationRequestStateStatus.CREATED, timestamp, lastUpdated: timestamp });
+    } catch (error: any) {
```
nklomp, line 37: The any is generating warnings

src/main/rp/ReplayRegistry.ts
```diff
@@ -0,0 +1,126 @@
...
+
+      const timestamp = Date.now();
+      const payload = await authorizationResponse.idToken.payload();
+      this.authorizationResponses.set(payload.nonce, {
+        payload,
+        status: AuthorizationResponseStateStatus.RECEIVED,
+        timestamp: timestamp,
+        lastUpdated: timestamp,
+      });
+    } catch (error: any) {
```
nklomp, line 56: The any is generating warnings

src/main/rp/ReplayRegistry.ts
```diff
@@ -0,0 +1,126 @@
...
+      this.authorizationRequests.set(payload.nonce, { payload, status: AuthorizationRequestStateStatus.CREATED, timestamp, lastUpdated: timestamp });
+    } catch (error: any) {
+      // TODO VDX-166 handle error
+    }
+  }
+
+  private async onAuthorizationResponseReceivedSuccess(authorizationResponse: AuthorizationResponse): Promise<void> {
+    try {
+      if (!authorizationResponse) {
+        return;
```
nklomp, line 45: throw an error. A success message needs a payload

src/main/rp/RP.ts
```diff
@@ -98,9 +123,22 @@ export class RP {
...
+      rpEventEmitter.emit(AuthorizationEvents.ON_AUTH_RESPONSE_RECEIVED_FAILED, authorizationResponsePayload, error);
```
nklomp, line 127: I would expect event objects, which then already create the timestamp of the event.

So create a generic event object with a subject that is a generic. Then in this example set the responsePayload as the subject. The event uses the current date when it is created. Then for handlers it means they do not have to generate timestamps. Also the event would have the error placeholder. So you can always have a method on the event hasError for instance

nklomp (changes requested):
I would have expected an event class with a subject/source in there. Which would take care of the timestamps and next to the payloads would contain any errors. That makes it easier for anyone else implementing listeners, as they have a contract for the events, instead of having to know per emission how to handle the listener.

Also I would move the emitters to the RP and OP itself. As you can have multiple instances of the classes. So containing a single emitter to an RP and OP is important.

Lastly the registry should probably have the emitter as constructor argument, to not bind it to a single shared emitter

src/main/rp/RP.ts
BtencateSphereon: cannpt really do that here as we need authorizationResponse later below

src/main/rp/ReplayRegistry.ts
BtencateSphereon: made some changes

src/main/rp/ReplayRegistry.ts
BtencateSphereon: added an error

src/main/rp/ReplayRegistry.ts
BtencateSphereon: made it an unknown, as it is not an Error

src/main/rp/ReplayRegistry.ts
BtencateSphereon: made it an unknown

src/main/rp/ReplayRegistry.ts
BtencateSphereon: added error

src/main/rp/RP.ts
BtencateSphereon: moved

src/main/op/OP.ts
BtencateSphereon: made some changes

src/main/op/OP.ts
BtencateSphereon: changed

src/main/op/OP.ts
```diff
@@ -59,8 +66,37 @@ export class OP {
     requestJwtOrUri: string | URI,
     requestOpts?: { nonce?: string; verification?: InternalVerification | ExternalVerification }
   ): Promise<VerifiedAuthorizationRequest> {
-    const authorizationRequest = await AuthorizationRequest.fromUriOrJwt(requestJwtOrUri);
-    return await authorizationRequest.verify(this.newVerifyAuthorizationRequestOpts({ ...requestOpts }));
+    const authorizationRequest = await AuthorizationRequest.fromUriOrJwt(requestJwtOrUri)
+      .then((result: AuthorizationRequest) => {
+        if (this._eventEmitter) {
```
nklomp, line 71: suggest to make a private method, that does the if check and emits if present. Then you get some nicer code instead of the 4 times `if (this._eventEmmiter)`

It could even retrhow the error if present. Meaning your code will always simply call emitEvent

So the body is doing something like this:

````typescript
ìf (this._eventEmmitter) {
   this._eventEmitter.emit(event)
}
if (event.hasError()) {
  throw event.getError()
}
````

nklomp (changes requested):
Yes this is better. One thing. Suggest to create an emitEvent method that does the logic you are now repeating over and over. That makes it way easier on the eyes and removes chances of introducing future bugs, because people might forget that the same expressions are scattered across the code

nklomp (approved):
LGTM

PR #58, opened 2023-06-22
fix: claims are not required

berendsliedrecht:
I was using the library for the most simple did auth and it seems like the `claims` field it not required by the specification. However, it was difficult for me to find the exact answer to this as the spec did not seem to formally define this. In the data model, within this library, `claims` may be undefined so this would be inline with that.


nklomp:
Claims are indeed not required. Because of a bug we ran into the same issue recently and it was fixed differently. Will leave this open to verify later this week, but should be resolved. Thanks for the PR

nklomp (dismissed):
LGTM Thanks

nklomp (approved):

PR #59, opened 2023-09-08
fix: alignment with openid.core and draft 11

TimoGlastra:
Some small adjustments to align the implementation with OpenID.Core and SIOP v2 Draft 11. Namely:
- Allow the `state` property to be omitted from the request. This is a recommended property, but not required. It now includes the state in the response when it was present in the request
- Update the IDToken method to extract the client metadata from verified authorization request, so we don't have to deal with registration (draft 11) and client_metadata (draft 11).

All tests pass, and we have tested this to work with a custom SIOPv2 Draft 11 implementation (as requester), and this lib as the responder.


nklomp (approved):
LGTM. Thanks

PR #60, opened 2023-09-28
fix: parse scheme with number

TimoGlastra:
Allows to parse schemes with numbers (e.g. `openid4vp://`).

Was causing a few issues. 

Made the regex a bit more complex, as the first chracater is techincally only allowed to be alpha character, and the following can also be digit and a few other characters


nklomp (approved):
LGTM, thanks

PR #61, opened 2023-09-28
chore: some changes for JFF interop

TimoGlastra:
Have been tinkering a bit to get interop with MATTR launchpad.

It  now seems to get stuck on the presentation definition validation, so will have to look at the PEX library for that.

Not sure if the JWT verification bypass is 'ok'. If there's no JWT I think it would be fine to just not validate it, but if there is one, it is validated.

Not ready to be merged, but if you do take a look, maybe these changes can  be of help for getting the sphereon wallet working...

### DIFF
--- a/src/authorization-request/AuthorizationRequest.ts
+++ b/src/authorization-request/AuthorizationRequest.ts
@@ -156,6 +156,10 @@ export class AuthorizationRequest {
       // TODO: We need to do something with the metadata probably
     }
     await checkWellknownDIDFromRequest(mergedPayload, opts);
+
+    // TODO: we need to verify somewhere that if response_mode is direct_post, that the response_uri may be present,
+    // BUT not both redirect_uri and response_uri. What is the best place to do this?
+
     const presentationDefinitions = await PresentationExchange.findValidPresentationDefinitions(mergedPayload, await this.getSupportedVersion());
     return {
       ...verifiedJwt,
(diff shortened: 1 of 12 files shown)


nklomp (commented):
Cool, I was going to work on this tomorrow

Be aware that at some points we might need to include some logic for version discovery. Because basically `direct_post` is replacing `post` from versions < 17. In the old `post` mode, redirect_uri was required to be present, whilst now it can also have the response_uri, but not both.

TimoGlastra:
Awesome thanks! I wasn't aware that it replaces post. More version logic... 😅

nklomp:
Yeah, long live evolving specs. I cannot wait once things settle down to start refactoring and remove a lot of the complexities from handling multiple versions..

BTW be aware that some changes have been made to this lib, mainly bugfixes

TimoGlastra:
Got it working with this very hacky encoding logic. Not sure how you're on adding dependencies, but with `qs` it's a one liner :) 

https://www.npmjs.com/package/qs?activeTab=dependencies

nklomp:
FYI: I am taking some of the changes from this PR in the branch I am working on for V18 support

TimoGlastra:
Yes, I was figuring that it's easiest if I push my 'hacks' and you can properly integrate it where needed. It's quite a mess in how I got things working :/

PR #64, opened 2023-12-23
feat: initial sd-jwt support

TimoGlastra:
Adds support for SD-JWT to be used as VPs.

The SdJwt.spec.ts has an end to end test case where a response is created based on a sd-jwt vc. 

It just adds support for the SD-JWT credential format profile, it hasn't implemented all restrictions from HAIP.

Dependant on https://github.com/Sphereon-Opensource/PEX/pull/132 to be merged and released first

### DIFF
--- a/package.json
+++ b/package.json
@@ -32,7 +32,7 @@
     "qs": "^6.11.2",
     "@sphereon/pex": "2.2.2",
     "@sphereon/pex-models": "^2.1.2",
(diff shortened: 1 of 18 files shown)


nklomp:
This has been merged and released as 0.6.0-unstable.0, which includes PEX 3.0.0, which in turn contains the SD-JWT support and some other small changes

PR #75, opened 2024-02-28
MOSIP fixes

sanderPostma:
include presentationSubmission in auth response
include nonce in AuthorizationResponsePayload


src/authorization-response/AuthorizationResponse.ts
```diff
@@ -120,7 +120,9 @@ export class AuthorizationResponse {
         presentationDefinitions,
         presentations: wrappedPresentations,
         verificationCallback: verifyOpts.verification.presentationVerificationCallback,
-        opts: { ...responseOpts.presentationExchange, hasher: verifyOpts.hasher },
+        opts: { ...responseOpts.presentationExchange,
+          presentationSubmission: (responseOpts.presentationExchange.presentationSubmission ?? authorizationResponsePayload.presentation_submission),
```
sanderPostma, line 124: assertValidVerifiablePresentations has this check
  ```
  } else if (args.presentationDefinitions && !args.opts.presentationSubmission) {
    throw new Error(`No presentation submission present. Please use presentationSubmission opt argument!`);
  }
  ```
  So we need opts.presentationSubmission

src/authorization-response/Payload.ts
```diff
@@ -19,11 +19,13 @@ export const createResponsePayload = async (
 
   // If state was in request, it must be in response
   const state: string | undefined = await authorizationRequest.getMergedProperty('state');
+  const payload = await authorizationRequest.requestObject.getPayload();
 
   const responsePayload: AuthorizationResponsePayload = {
     ...(responseOpts.accessToken && { access_token: responseOpts.accessToken }),
     ...(responseOpts.tokenType && { token_type: responseOpts.tokenType }),
     ...(responseOpts.refreshToken && { refresh_token: responseOpts.refreshToken }),
+    ...(payload?.nonce && { nonce: payload.nonce}),
```
sanderPostma, line 28: In AuthenticationResponse verify():
```
    const nonce = merged.nonce ?? verifiedIdToken?.payload.nonce ?? oid4vp.nonce;
    ...
    } else if (oid4vp.presentationDefinitions.length > 0 && !nonce) {
      throw Error('Nonce is required when using OID4VP');
    }
```
In my case, if I don't return the nonce it will run into "Nonce is required when using OID4VP"

src/authorization-response/AuthorizationResponse.ts
```diff
@@ -115,14 +115,16 @@ export class AuthorizationResponse {
...
+        responseOpts.presentationExchange?.presentationSubmission ?? authorizationResponsePayload.presentation_submission;
 
       await assertValidVerifiablePresentations({
         presentationDefinitions,
         presentations: wrappedPresentations,
         verificationCallback: verifyOpts.verification.presentationVerificationCallback,
         opts: {
           ...responseOpts.presentationExchange,
-          presentationSubmission: responseOpts.presentationExchange?.presentationSubmission ?? authorizationResponsePayload.presentation_submission,
+          ...(presentationSubmission ? { presentationSubmission: presentationSubmission } : {}),
```
nklomp, line 127: This is wrong! If the above hasVpToken is true, there needs to be a submission. So either the logic to determine the hasVpToken is wrong, or something else is happening. This PR is wrong for sure

nklomp (changes requested):
PR is wrong. if there is a vp token in the request, you are required to have a presentation submission

src/authorization-response/Payload.ts
```diff
@@ -19,11 +19,13 @@ export const createResponsePayload = async (
 
   // If state was in request, it must be in response
   const state: string | undefined = await authorizationRequest.getMergedProperty('state');
+  const payload = await authorizationRequest.requestObject?.getPayload();
 
   const responsePayload: AuthorizationResponsePayload = {
     ...(responseOpts.accessToken && { access_token: responseOpts.accessToken }),
     ...(responseOpts.tokenType && { token_type: responseOpts.tokenType }),
     ...(responseOpts.refreshToken && { refresh_token: responseOpts.refreshToken }),
+    ...(payload?.nonce && { nonce: payload.nonce }),
```
nklomp, line 28: Nonce should not endup in authorization response. That should be either in the id token or vp token

src/authorization-response/AuthorizationResponse.ts
sanderPostma: I found the submission in authorizationResponsePayload.presentation_submission not in presentationExchange. So this needs to be fixed elsewhere then.
So that that makes this PR is not to merge, just for diagnostics. I will revert the change in my branch as soon as presentationSubmission appears in presentationExchange

src/authorization-response/Payload.ts
sanderPostma:
```
const oid4vp = await verifyPresentations(this, verifyOpts);
const nonce = merged.nonce ?? verifiedIdToken?.payload.nonce ?? oid4vp.nonce;
```
When I remove nonce from AuthorizationResponsePayload  it will not be in merged.nonce, I don't have idToken so I will need to debug verifyPresentations .

src/authorization-response/AuthorizationResponse.ts
nklomp: No you are likely simply misunderstanding how it works. Presentation Exchange is a service that provides the submission,  but only when a vp_token is needed

src/authorization-response/Payload.ts
nklomp: As mentioned it needs to end either in the id token and/or vp token depending om which token(s) are requested. Since this comes from the OID request object it should never end up in oauth2 response payload directly

nklomp: So if there is no request object then there indeed is no nonce typically

src/authorization-response/AuthorizationResponse.ts
sanderPostma: I have found the problem, in the SSI SDK version I am using with RSA support, presentationSubmission in OPSession is not loaded yet.  But this line still has to be in opts
```
          ...(presentationSubmission ? { presentationSubmission: presentationSubmission } : {}),
```
It's asserted in assertValidVerifiablePresentations
```
  } else if (args.presentationDefinitions && !args.opts.presentationSubmission) {
    throw new Error(`No presentation submission present. Please use presentationSubmission opt argument!`);
  } else if (args.presentationDefinitions && presentationsWithFormat) {

```

nklomp: No that line should not be in opts

sanderPostma: Ah I see now it has the ... before responseOpts.presentationExchange so it should be in opts already.  So the problem was my OPSession which was from before October 13th and I patched before it up by getting it from elsewhere.
I reverted that code

src/authorization-response/Payload.ts
sanderPostma: I reverted this. After retest with newer libraries this appears to be working again.

sanderPostma:
Closing this PR, the left hand develop branch does not match what's actually in that branch atm

PR #76, opened 2024-04-11
fix: always verify nonce, extract nonce from VP

TimoGlastra:
This PR makes a few improvements to the nonce extraction and verification process:
- correctly extract nonce from SD-JWT, W3C JSON-LD or W3C JWT presentation. (I'll make a small follow up pr to clean up the SD-JWT nonce a bit once https://github.com/Sphereon-Opensource/SSI-SDK/pull/183 is merged and released)
- Fix the extraction of nonce values from set using array index (always returns undefined)
- Make sure all presentations and id token and auth response use the same nonce
- make sure nonce is always defined
- check nonce against verifyOpts if provided


nklomp (approved):
LGTM

PR #77, opened 2024-04-13
fix: no async in filter

TimoGlastra:
This one took me really long to debug, but i was finally able to find out why a succesful verification event was emitted even though the presentation verification callback failed. 

The filter that had an `async` method doesn't really work, and thus it would see the returned promise value in the filter method as a truthy value (`Boolean(Promise.reject())` will still evaluate to true). It was especially weird since the error was thrown, but because it was not handled within a promise, it was an unhandled rejection.

It does mean that the presentation verification callback has had no effect on the outcome of the verification, so this seems like a serious bug to get released as soon as possible.


nklomp (approved):
LGTM

PR #78, opened 2024-04-22
Updating to OID4VP_1_0_20

sksadjad:
updated in this pr:
- added three new client id scheme:
  - verifier_attestation
  - x509_san_dns
  - x509_san_uri
implemented the checks for verifier_attestation and san_dns


generator/schemaGenerator.ts
```diff
@@ -165,20 +165,20 @@ const authorizationRequestPayloadVD11 = {
...
+  schemaId: 'AuthorizationRequestPayloadVD13OID4VPD20Schema',
+  outputPath: 'src/schemas/AuthorizationRequestPayloadVD13OID4VPD20.schema.ts',
   // outputConstName: 'AuthorizationRequestPayloadSchemaVD11',
   skipTypeCheck: true
 };
 
 let schemas: Schema[] = [
   writeSchema(authorizationRequestPayloadVID1),
   writeSchema(authorizationRequestPayloadVD11),
-  writeSchema(authorizationRequestPayloadVD12OID4VPD18),
```
nklomp, line 181: Why are we removing v18?

src/authorization-request/AuthorizationRequest.ts
```diff
@@ -1,4 +1,7 @@
 import { JWTVerifyOptions } from 'did-jwt';
+import { decodeJWT } from 'did-jwt';
+import { JWTDecoded } from 'did-jwt/lib/JWT';
+import forge from 'node-forge';
```
nklomp, line 4: This cannot work. This library is used in browser, RN and Node. The lib is node only

src/helpers/SIOPSpecVersion.ts
```diff
@@ -34,15 +34,15 @@ export const authorizationRequestVersionDiscovery = (authorizationRequest: Autho
   const versions = [];
   const authorizationRequestCopy: AuthorizationRequestPayload = JSON.parse(JSON.stringify(authorizationRequest));
   // todo: We could use v11 validation for v12 for now, as we do not differentiate in the schema at this point\
-  const vd12Validation = AuthorizationRequestPayloadVD12OID4VPD18Schema(authorizationRequestCopy);
```
nklomp, line 37: Don't remove older versions

TimoGlastra:
It seems there's some overlap with this PR and #82 with the new client_id_schemes. However in #82 the x509 certificate needs to be verified by the JWT verifier callback. We thought this helps in keeping this library a bit more lightweight (and we don't have to import a x509 library)

PR #82, opened 2024-06-30
feat: remove did-resolver and did-jwt dependencies

auer-martin:
This PR mainly removes the strong dependencies to the did-resolver and did-jwt dependencies and allows consumers to provide signing and verification mechanisms. This was discussed before in Sphereon-Opensource/SIOP-OID4VP#55.

Some notable changes:

The Signing and Verification Capabilities now have to be provided externally. Thus, internal checks, such as verifying linked domains, were removed and must now be done as part of the pluggable verification process.

The different signature types INTERNAL | SUPPLIED | ... were removed.
The behavior can be mimicked by implementing the signing callback accordingly.

The Verification Method types INTERNAL | EXTERNAL were removed.
Users now must provide their own jwt verification implementation. The INTERNAL and EXTERNAL functionality
again can be achieved by implementing the verification callback accordingly.

This PR provides verification and signing 'adapters' for x5c, jwk, and did protected jwts (x5c, and jwk functionality was not present/possible previously) to make the user-provided signing and verification process as easy as possible. For example, when creating an id-token that is jwk protected, you can pass a jwtIssuer with method jwk, and the jwt header and payload are prepared so that the consumer of the library should not need to modify the jwt header/payload content manually. For verification, the JwtVerifier provides properties to streamline the verification process.

With these changes, I believe this PR effectively addresses the issues raised in Sphereon-Opensource/OID4VC#154 and the discussions in Sphereon-Opensource/SIOP-OID4VP#55.


src/types/JwtVerifier.ts
```diff
@@ -0,0 +1,91 @@
...
+import { JWK, JwtHeader, JwtPayload } from './JWT.types';
+
+export type JwtVerificationType = 'id-token' | 'request-object';
+
+interface DidJwtVerifier {
+  method: 'did';
+  type: JwtVerificationType;
+
+  didUrl: string;
+}
```
TimoGlastra, line 13: For type inference it may be easier to have  base type where they all extend from:



```
interface JwtVerifierBase {
 method: string
 jwtType: JwtVerificationType
}

Also maybe rename type to jwtType?

**Review by TimoGlastra — APPROVED** [review]
Nice 👍

File: src/helpers/jwtUtils.ts [hunk 2]
```diff
@@ -0,0 +1,34 @@
...
+    let did;
+    if (payload.sub && payload.sub.startsWith('did:')) {
+      did = payload.sub;
+    }
+    if (!did && header && header.kid && header.kid.startsWith('did:')) {
+      did = header.kid.split('#')[0];
+    }
+    if (did) {
+      return did;
+    }
```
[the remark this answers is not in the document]
auer-martin: I think we can remove this. I initially tried to preserve the behaviour here, but it seems like it's not really used anywhere in this library, and if someone needs the information, they can always extract it from the jwt itself.

src/types/JwtIssuer.ts
```diff
@@ -0,0 +1,56 @@
...
+interface JwtIssuerX5c {
+  method: 'x5c';
+
+  /**
+   *
+   * Array of base64-encoded certificate strings in the DER-format.
+   *
+   * The certificate containing the public key corresponding to the key used to digitally sign the JWS MUST be the first certificate.
+   */
+  chain: Array<string>;
```
nklomp, line 24: suggest to simply call this x5c as that is what the contents is

src/types/JwtIssuer.ts
```diff
@@ -0,0 +1,56 @@
...
+  method: 'jwk';
+
+  jwk: JsonWebKey;
+}
+
+interface JwtIssuerCustom extends Record<string, unknown> {
+  method: 'custom';
+}
+
+export type JwtIssuer = JwtIssuerDid | JwtIssuerX5c | JwtIssuerJwk | JwtIssuerCustom;
```
nklomp, line 50: Given you are using interfaces, I would rather see the method property extracted into a type as well

test/AuthenticationRequest.request.spec.ts
```diff
@@ -269,88 +279,44 @@ describe('create Request Uri should', () => {
...
+        jwtIssuer: { method: 'did', didUrl: KID, alg: SigningAlgo.ES256K },
+        passBy: 'other type' as never,
+        createJwtCallback: getCreateJwtCallback({
           hexPrivateKey: HEX_KEY,
           did: DID,
           kid: KID,
-        },
-      },
-      registration: {
-        idTokenSigningAlgValuesSupported: [SigningAlgo.EDDSA, SigningAlgo.ES256],
```
nklomp, line 288: Why has this object been removed?

test/AuthenticationRequest.request.spec.ts
```diff
@@ -269,88 +279,44 @@ describe('create Request Uri should', () => {
...
-
-        signature: {
+        jwtIssuer: { method: 'did', didUrl: KID, alg: SigningAlgo.ES256K },
+        createJwtCallback: getCreateJwtCallback({
           hexPrivateKey: HEX_KEY,
           did: DID,
           kid: KID,
-        },
-      },
-      registration: {
```
nklomp, line 317: Why has registration been removed?

nklomp (changes requested):
I like the general direction and implementation of the PR. 🥇 

A few notes though:
- When working on this large of a PR next time probably it is best to have a small chat/discussion up front. To ensure direction is alligned, but also to ensure people are not by accident doing similar work

- I think we should create another module with the DID and well-known DIDs, as the library now only shows how to implement it using tests. This means we lost Microsoft Entra Verified ID compatibility, universal resolution and DIDs in general for parties that are not using a framework

- I see some expressions without accolades. We always require them to prevent any future potential bugs

- I think we lost the support for OAuth2 style payloads only, or to convert these into OID request object style. Although most examples in the specs are using request objects there are cases where the spec also mentions to not use signed JWTs. I have seen some examples where an async jwt() method is replaced by the _jwt property, and also some cases where a authz payload was converted to the request object have been removed on first glance

TimoGlastra:
Hey @nklomp, thanks for the feedback. I discussed with Martin all the changes, and we discussed briefly about opening an issue, but as it was already partly discussed in the other open issues we in the end didn't. 

I feel like we've discussed in the past quite often about the general purpose of these libraries, which is to provide an unopiniated foundation of the specification. To achieve that we need to be flexible, and so it made sense to us to remove the did specific logic. We looked at OID4VCI as an example, which doesn't provide did resolving logic as well. 

Nowhere in the OID4VP or SIOPv2 specification I see anything mentioned about DID Linked Domains. Would you prefer this is exctracted to a separate library in ssi-sdk? Or maybe we can in the JWT verification callback, optionally allow a did document to be returned like is done in OID4VCI, then we can do the verification there?

auer-martin:
Hello @nklomp, thanks for the feedback 👍🏻 , and apologies on my part for not opening an issue in advance. I will do so next time.
I will look into your feedback today and wait for a decision on how to proceed with the DID related functionality.

nklomp:
> Hey @nklomp, thanks for the feedback. I discussed with Martin all the changes, and we discussed briefly about opening an issue, but as it was already partly discussed in the other open issues we in the end didn't.
> 
> I feel like we've discussed in the past quite often about the general purpose of these libraries, which is to provide an unopiniated foundation of the specification. To achieve that we need to be flexible, and so it made sense to us to remove the did specific logic. We looked at OID4VCI as an example, which doesn't provide did resolving logic as well.
> 
> Nowhere in the OID4VP or SIOPv2 specification I see anything mentioned about DID Linked Domains. Would you prefer this is exctracted to a separate library in ssi-sdk? Or maybe we can in the JWT verification callback, optionally allow a did document to be returned like is done in OID4VCI, then we can do the verification there?

Yeah, point is more that this PR removed existing functionality. I rather would have seen that discussed upfont, so we could have discussed how to go about it.

As I already mentioned to you, what I want to do soon is to basically merge the current SIOP/OID4VP lib and the OID4VCI lib into a mono repo. Then to also split up SIOP and OID4VP more into separate modules and move a lot of the old version support. I also already want to make an abstraction for query language support for VP, so that we can plugin PEx and any future spec for it.

For now the easiest approach is to convert this repo already into a mono-repo. Then to add a module that supports DIDs including the universal resolver/registrar lib as that only depends on cross-fetch. Ideally well-known DIDs would be a separate module, but I would be okay with having it in the DID module for a first version.

Well-known DIDs are part of the JWT VC Presentation Interop profile which the lib supports. That DIF profile is also supported by different larger tech vendors, including Microsoft Entra Verified ID.

To conclude. I am happy with the approach as mentioned and this indeed is the route forward. But next time I would like to see some discussion upfront in case we will have to (partially) drop certain features present, so we can come up with a plan and decide who will do what at which specific time :)

src/id-token/Payload.ts
```diff
@@ -50,34 +37,11 @@ export const createIDTokenPayload = async (
     aud: responseOpts.audience || payload.client_id,
     iat: Math.round(Date.now() / SEC_IN_MS - 60 * SEC_IN_MS),
     exp: Math.round(Date.now() / SEC_IN_MS + (responseOpts.expiresIn || 600)),
-    sub: responseOpts.signature.did,
+    sub: undefined,
```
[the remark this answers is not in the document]
auer-martin: So when you build an idtoken with the `fromVerifiedAuthorizationRequest` you dont' have the jwtIssuer available.
Thus you don't know the subject yet. I set this when calling the jwt() method on the idToken

src/authorization-request/URI.ts
```diff
@@ -151,15 +152,13 @@ export class URI implements AuthorizationRequestURI {
     }
 
     const isJwt = typeof authorizationRequestPayload === 'string';
-    const requestObjectJwt = requestObject
-      ? await requestObject.toJwt()
-      : typeof authorizationRequestPayload === 'string'
```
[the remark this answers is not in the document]
auer-martin: I changed this as it was before. However, removing it should have been okay since the AuthorizationRequestPayload cannot be a string based on its type.

src/id-token/IDToken.ts
```diff
@@ -111,30 +151,23 @@ export class IDToken {
   public async verify(verifyOpts: VerifyAuthorizationResponseOpts): Promise<VerifiedIDToken> {
     assertValidVerifyOpts(verifyOpts);
 
-    const { header, payload } = parseJWT(await this.jwt());
-    this.assertValidResponseJWT({ header, payload });
+    const parsedJwt = parseJWT(this._jwt);
```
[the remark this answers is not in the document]
auer-martin: Okay, so here's the problem with leaving it as it was.
The method this.jwt() creates the jwt if it hasn't been created yet, which I find strange because that means we're verifying a jwt that we previously created. We can keep the current behaviour, but then how do we create the jwt? Do we sign it with the custom method? Passing a JwtIssuer to the verification method?
For now I changed the implementation that it throws an Error if this._jwt is not set. Not sure what you would prefer here?

TimoGlastra:
> Yeah, point is more that this PR removed existing functionality. I rather would have seen that discussed upfont, so we could have discussed how to go about it.

Yeah our bad, we'll discuss it upfront next time -- 
The burden is on us to fix it now 😉

> For now the easiest approach is to convert this repo already into a mono-repo. Then to add a module that supports DIDs including the universal resolver/registrar lib as that only depends on cross-fetch. Ideally well-known DIDs would be a separate module, but I would be okay with having it in the DID module for a first version.

There's quite some work and overhead involved in moving a repository to a monorepo, including the release pipeline etc.. I think my preference would be to move this siop-oid4vp and the new did module already to the oid4vci monorepo for example and then later we can split it up. I think there's not too many PRs open now, so it should be quite straightforward. And my first moving this repo as is, means we have the code in one repo, without the overhead of separating modules (we can start doing that work over time). Are you okay with that approach?

In that case, it'd be good to:
- we address all comments on this PR except the DID package
- we merge this PR
- we move the code over to oid4vci repo (and rename the repo to oid4vc) as-is as a new package
- we add the new did related package in a new PR to the oid4vc repo

This way the effort spent will be on the desired solution (one monorepo) instead of spending effort to transform this into a monorepo.

src/authorization-request/types.ts
```diff
@@ -34,6 +35,7 @@ export interface RequestObjectPayloadOpts<CT extends ClaimPayloadCommonOpts> {
   scope: string; // from openid-connect-self-issued-v2-1_0-ID1
   response_type: string; // from openid-connect-self-issued-v2-1_0-ID1
   client_id: string; // from openid-connect-self-issued-v2-1_0-ID1
+  client_id_scheme: ClientIdScheme;
```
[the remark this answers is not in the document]
auer-martin: you are right, sorry for that

README.md
```diff
@@ -263,45 +259,65 @@ You could also use the actual example keys and DIDs, as they are valid Ethr Rops
...
+      // verify jwk certificate protected jwt's
+    } else if (jwtVerifier.method === 'custom') {
+      // Only called if based on the jwt the verification method could not be determined 
+      throw new Error(`Unsupported JWT verifier method ${jwtIssuer.method}`)
+    }
+  }
+}
+
+function createJwtCallback(): CreateJwtCallback {
+  return async (jwtIssuer, jwt) => {
```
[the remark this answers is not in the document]
auer-martin: Okay, so I fixed it this way: the jwtIssuer now takes an additional optional argument called `option`, which is just a `Record<string, unknown>` where you can pass additional context such as a kid.

test/AuthenticationRequest.request.spec.ts
auer-martin: added again

test/AuthenticationRequest.request.spec.ts
auer-martin: added again

src/id-token/IDToken.ts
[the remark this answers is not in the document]
nklomp: The whole idea is to do a lazy evaluation. Be aware that the payloads can be used both on the holder side as well as on the RP side. So whenever you already had a JWT string, you could use the class to pass that in and get the accompanying decoded payload. But as a RP you could also build the request object first and then convert it into a JWT

auer-martin:
Hey @nklomp is there anything else that needs to happen before we can merge this?

src/id-token/Payload.ts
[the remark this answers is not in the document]
nklomp: Yes, but why is the sub key/prop being included in that case to begin with?

src/types/JwtVerifier.ts
```diff
@@ -0,0 +1,93 @@
...
+interface X5cJwtVerifier extends JwtVerifierBase {
+  method: 'x5c';
+
+  /**
+   *
+   * Array of base64-encoded certificate strings in the DER-format.
+   *
+   * The certificate containing the public key corresponding to the key used to digitally sign the JWS MUST be the first certificate.
+   */
+  chain: Array<string>;
```
nklomp, line 26: x5c like in the Jwt issuer

nklomp (changes requested):
A few small comments left.

BTW we prefer to have the reviewer resolve the raised issues. Reason is that it ensures the reviewer actually has seen the change implemented and is okay with it. Otherwise it is easy to overlook and/or harder to find whether it was resolved appropriately

src/types/JwtVerifier.ts
auer-martin: done

nklomp (approved):

nklomp:
Okay, there are still some test failures. I guess some are related to ehtr:goerly not being around anymore, similar for ebsi testnetwork not being around anymore

TimoGlastra:
Do you want us to go ahead with moving the siop-oid4vp repo to the OID4VCI monorepo as well? I think we can change the package name at the same time (so it's not did-auth-siop anymore).

Or do you want to wait with that?

nklomp:
Sure. Suggest to call it siop-oid4vp for now though, as with the next refactor extracting SIOPv2 and OID4VP into respective modules is highly likely

nklomp:
To make it not too hard for people to have to adjust their imports all over the place, we could opt for leaving the name as is for now.

TimoGlastra:
> To make it not too hard for people to have to adjust their imports all over the place, we could opt for leaving the name as is for now.

Sure, then we can rename when we separate the modules 👍

auer-martin:
I created a PR to move this repo over to the OID4VC monorepo. https://github.com/Sphereon-Opensource/OID4VC/pull/125
I added some small changes compared to this PR, mainly adding support for v20 and client_id_schemes. The changes can be found [here](https://github.com/auer-martin/SIOP-OID4VP/commit/d0823e41a1e236fd970c5f212edb182621140c9c). 
In addition, I fixed the tests in the OID4VC repo. I skipped the ebsi tests, as the mock endpoint no longer exists, and I could not find a proper replacement on the ebsi docs.

nklomp:
Has been implemented in the new location

